PR #12, opened 2020-08-25
FIZZ-9 added Google Classroom diagram to the drawio file, moved to an…

ggaronzik:
FIZZ-9 added Google Classroom diagram to the drawio file, moved to an app-agnostic folder


bradbanister:
@ggaronzik On the Reports API for Google Classroom, I'm using UserUsageReport.timestamp_last_interaction for engagement/participation.

bradbanister:
Also, while the CustomerUsageReport num_1day_students and num_7day_students are a good find, i'm not sure what's the right category.  I don't think it's attendance in the traditional sense.  Maybe more of a class-wide engagement?

ejansson (approved):

PR #15, opened 2020-08-31
[FIZZ-21] Generate csv with a list of students with missing assignments

lalrg:


canvas/fizz-21-py/readme.md
```diff
@@ -0,0 +1,27 @@
+# Generate a list of users with missing assignments
+
+### What does it do?
+Input:
```
stephenfuqua, line 4: Markdown headings should have an empty line on each side of them.
```suggestion

Input:
```

canvas/fizz-21-py/readme.md
```diff
@@ -0,0 +1,27 @@
...
+CSV files with information about:
+* Courses
```
stephenfuqua, line 6: Markdown lists should also be surrounded by blank lines.
```suggestion

* Courses
```

canvas/fizz-21-py/readme.md
```diff
@@ -0,0 +1,27 @@
...
+* A CVS file with information about students with missing assignments.
+
+### How to use it
+1. Run the command ``pipenv install``.
+2. Copy the .env.example file to .env and update its values.
+3. Enable the corresponding virtual environment.
+4. Open and run the test.ipynb file
+
+### .env information
+CSV_INPUT_BASE_PATH=[PATH_WHERE_INPUT_CSV_WILL_BE_LOCATED]
```
stephenfuqua, line 22: I think you can omit this section on ".env information", since step 2 already tells the user to copy and update the file, and the file contents are self explanatory.

canvas/fizz-21-py/src/test.ipynb
```diff
@@ -0,0 +1,186 @@
...
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "from dotenv import load_dotenv\n",
+    "import os\n",
+    "\n",
+    "load_dotenv()\n",
+    "os.chdir(os.getenv(\"CSV_INPUT_BASE_PATH\"))\n",
```
stephenfuqua, line 51: As a general rule, I would put the full path into the `read_csv` command instead of changing the local working directory. When changing the working directory, it is best to restore to the original location. If the program crashes, then it might not get to the point of changing back to that original location. Thus best not to change locations.

canvas/fizz-21-py/.gitignore
```diff
@@ -0,0 +1,145 @@
+# Originated from https://github.com/github/gitignore/blob/master/Python.gitignore
```
stephenfuqua, line 1: Can you go ahead and load most of this into the root `.gitignore` file? Except for `.vscode/`, because generally we *do* want to share these settings across users.

PR #17, opened 2020-09-01
[FIZZ-27] Data extraction with python

lalrg:


canvas/fizz-27-py/readme.md
```diff
@@ -0,0 +1,29 @@
...
+* Assignments
+* Students
+* Enrollments
+* Submissions
+
+### Requirements
+
+* python 3.8
+* pip 20.2.2
+* pipenv 2020.8.13
```
bradbanister, line 22: We should all discuss moving from pipenv to Poetry. It was an easy move and has been a much better experience for me. It also sets the stage for easy publication to Azure Artifacts.

canvas/fizz-27-py/src/lib/canvasRequest.py
```diff
@@ -0,0 +1,10 @@
+import requests
+import os
+
+def get(url):
```
bradbanister, line 4: Consider adding retries here. I'm using https://github.com/channable/opnieuw and it's quite easy to mix in. Just a function decorator.

canvas/fizz-27-py/src/lib/assignments.py
```diff
@@ -0,0 +1,10 @@
+from lib.canvasRequest import get
+
+def getAssignments(courseId, assignmentStatus):
+    assignments = get(f"/api/v1/courses/{courseId}/assignments?bucket={assignmentStatus}")
+    return assignments
+
```
bradbanister, line 6: You should set up Black to auto format your code:  https://github.com/psf/black.  For example, Python coding conventions apparently want two lines between function declarations.

canvas/fizz-27-py/src/lib/canvasRequest.py
stephenfuqua: Retries is an interesting point and might make it worthwhile to fork the [canvasapi](https://github.com/ucfopen/canvasapi) package or continue with our own code, narrowly scoped to what we need.

One of the benefits of `canvasapi` is that it has a large test suite already. Also has built-in support for working with pagination.

canvas/fizz-27-py/src/data-extractor.py
```diff
@@ -0,0 +1,40 @@
...
+from lib.assignments import getAssignmentsForCourse
+from lib.submissions import getSubmissionsForAssignment
+
+load_dotenv()
+
+all_courses = getCoursesForYear()
+all_students = list()
+all_assignments = list()
+all_submissions = list()
+all_enrollments = list()
```
stephenfuqua, line 14: Tip: you can also use `all_enrollments = []`.

canvas/fizz-27-py/src/data-extractor.py
```diff
@@ -0,0 +1,40 @@
...
+    students = getStudentsForCourse(course_id)
+    assignments = getAssignmentsForCourse(course_id)
+
+    for assignment in assignments:
+        assignment_id = assignment["id"]
+        submissions_for_assignment = getSubmissionsForAssignment(course_id, assignment_id)
+        all_submissions = all_submissions + submissions_for_assignment
+
+    students = list(filter(lambda student: student not in all_students, students))
+    all_students = all_students + students
```
stephenfuqua, line 27: If `all_students` is intended not to have duplicates, then it may make sense to create a `set` instead of a `list`. This then simplifies lines 26 and 27 to `all_students = all_students | set(students)`.

canvas/fizz-27-py/src/lib/canvasRequest.py
```diff
@@ -0,0 +1,10 @@
...
+    base_url = os.getenv("CANVAS_BASE_URL")
+    auth_token = os.getenv("ACCESS_TOKEN")
```
stephenfuqua, line 6: Eventually, if we create our own library interface, we would want to pass these values into a constructor instead of pulling them from the environment variables here.

canvas/fizz-27-py/src/lib/canvasRequest.py
```diff
@@ -0,0 +1,10 @@
...
+    r = requests.get(
+        base_url + "/" + url,
+        headers={ "Authorization" : f"Bearer {auth_token}" },)
```
stephenfuqua, line 9: I have read several articles on strings in Python and not come across string interpolation with `f` before now! And yet it was added in Python 3.6, I now see, several years ago. Sweet. In all of the articles I had come across before they were still recommending `"Bearer {}".format(auth_token)`, which is an improvement over the older % formatting. I'm going to start using this `f` formatting style.

canvas/fizz-27-py/src/lib/courses.py
```diff
@@ -0,0 +1,33 @@
...
+    return getCoursesByStatus("available")
+
+def getCompletedCourses():
+    return getCoursesByStatus("completed")
+
+def getCoursesForYear():
+    available_courses = getAvailableCourses()
+    completed_courses = getCompletedCourses()
+    total_courses = available_courses + completed_courses
+    return list(filter(filterCoursesByYear, total_courses))
```
stephenfuqua, line 20: Suggestion to start using list comprehensions. Essentially the same as what you've done but considered "more pythonic".

```
return [tc for tc in total_courses if filterCoursesByYear(tc)]
```

If taking that approach, for readability I would change the function name to `course_in_report_year`.

canvas/fizz-27-py/src/lib/courses.py
```diff
@@ -0,0 +1,33 @@
+from lib.canvasRequest import get
+import os
+from datetime import datetime
+
+def getCoursesByStatus(status):
```
stephenfuqua, line 5: https://www.python.org/dev/peps/pep-0008/#id45:

> "Function names should be lowercase, with words separated by underscores as necessary to improve readability.
>
> Variable names follow the same convention as function names.
>
> mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility."

bradbanister: I didn't even notice this, I'm so much more use to non-snake case function names!  FYI, PyLint gives warnings about this.

PR #18, opened 2020-09-03
[FIZZ 26] Add paging to google extractor

bradbanister:


src/google/README.md
```diff
@@ -115,15 +117,9 @@ in general can be found
 
 To work in Visual Studio Code install the Python Extension.
 Then:
-1. Close Visual Studio Code and open a new command prompt in
-   the root directory of this project.
-1. Run `poetry shell` followed by `code .` (Don't forget the
-   period!  Note you will only need run Visual Studio Code like
-   this once.)
-1. Visual Studio Code will open. Type `Ctrl-Shift-P`, then
+1. Type `Ctrl-Shift-P`, then
```
lalrg, line 120: We could specify that this shortcut is specifically for windows/linux.

lalrg (approved):

src/google/google-extractor/main.py
```diff
@@ -34,62 +36,85 @@ def get_credentials():
...
+
+
+# resource_method is the get/list SDK function to call,
+# resource_parameters is the parameters for get/list
+# response_property is the property in the response json we want
+# results is the result accumulated across pages
+@tail_recursive
+def request_execution(
+    resource_method, resource_parameters, response_property, results=None
+):
```
stephenfuqua, line 55: There are several good styles to adopt for class and function documentation. Perhaps [numpydoc](https://numpydoc.readthedocs.io/en/latest/) would be a good one to choose.

```suggestion
@tail_recursive
def request_execution(
    resource_method, resource_parameters, response_property, results=None
):
    """ Execute a REST API request
   
    Parameters:
    ----------
    resource_method: list
        is the get/list SDK function to call
    resource_parameters: list
        is the parameters for get/list
    response_property: string
        is the property in the response json we want
    results: list
        is the result accumulated across pages   
    """
```

Adjust the data types if I interpreted incorrectly :-).

src/google/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "google-extractor"
-version = "0.1.1"
+version = "0.1.2"
 description = ""
 authors = ["Brad Banister <brad@doublelinepartners.com>"]
```
stephenfuqua, line 5: In other projects we generally have authorship as `Ed-Fi Alliance, LLC, and contributors`. Wonder if that would make sense here?

stephenfuqua: Also, please add: `license = Apache-2.0`

stephenfuqua (approved):

src/google/google-extractor/main.py
bradbanister: Thanks, this was definitely the next step here.

src/google/pyproject.toml
bradbanister: Sounds good.  Author was auto-generated.

PR #19, opened 2020-09-03
[FIZZ-35] Change folder structure and add readme file for Canvas spikes and stories

lalrg:
We could create another story in order to consolidate the virtual environments, having a common pipfile, .env file, .pylintrc and so on.


bradbanister:
Looking at everything, I don't think it's very end-user friendly the way it is partitioned by FIZZ ticket.  I'd suggest either going ahead and pulling things together into a single project, or at least naming the projects and folders after their purpose.

Not required now, but I think the next release needs to include project versioning in a pyproject.toml file (newer standard used by Poetry and Flit).

lalrg:
> Looking at everything, I don't think it's very end-user friendly the way it is partitioned by FIZZ ticket. I'd suggest either going ahead and pulling things together into a single project, or at least naming the projects and folders after their purpose.
> 
> Not required now, but I think the next release needs to include project versioning in a pyproject.toml file (newer standard used by Poetry and Flit).

I think we can change the name of the folders, and later we could add another story for unifying the folders in a single project

stephenfuqua:
> > Looking at everything, I don't think it's very end-user friendly the way it is partitioned by FIZZ ticket. I'd suggest either going ahead and pulling things together into a single project, or at least naming the projects and folders after their purpose.
> > Not required now, but I think the next release needs to include project versioning in a pyproject.toml file (newer standard used by Poetry and Flit).
> 
> I think we can change the name of the folders, and later we could add another story for unifying the folders in a single project

I was picturing the released work being well organized, but not too concerned about somewhat disorganized experimental work. But the experimental work *should* be re-arranged a bit sometime soon.

src/canvas/.gitignore
```diff
@@ -1,2 +1,3 @@
 # VS Code
 .vscode/
+*.csv
```
stephenfuqua, line 3: Are you sure we need this file? Root level `.gitignore` already has `*.csv`, and generally we _don't_ want to ignore `.vscode` directories.

stephenfuqua (approved):
I didn't have time for a thorough review and don't want to hold this up. Go ahead please. Can always refactor a little more when/if needed. This is a helpful re-org.

src/canvas/.gitignore
bradbanister: @stephenfuqua  On un-ignore of .vscode, can only do if "python.pythonPath" is relative path.  pipenv is absolute path in user directory by default.  Poetry setting puts in a .venv folder in workspace, making path ".venv\\Scripts\\python.exe" work for everyone.

stephenfuqua: `"python.venvPath": "${workspaceFolder}/.venv/bin/python"` is what we have in the root `.vscode` folder.

bradbanister: After giving Poetry the "in project" config, it created a directory scheme with python.exe in .venv\Scripts.  VS Code picked saw it and added `"python.pythonPath":` ".venv\\Scripts\\python.exe",`

PR #20, opened 2020-09-03
[FIZZ-24] Schoology Assignment Data Extractor

stephenfuqua:
This is functional, although it doesn't meet my long-term quality or functional objectives:

1. I will likely strip out the `schoolopy` classes and start over on the Schoology API.
  * Better incorporate extra methods
  * Handle paging
  * Don't swallow errors
  * Build in retries
  * Normalize some data types: Schoology sometimes uses string in one place and int in another for the same value, e.g. a student `uid`. 
  * Inadequate readme.
2. No tests
3. Probably want to accept command line arguments instead of only supporting `.env` file.
4. Internal documentation needs improvement

... and more, I'm sure.


experimental/schoology/export-assignment-data.py
```diff
@@ -0,0 +1,269 @@
...
+                # is of the form {section_id : student_uid}.
+                enrollments = section["enrollments"]
+                assignees = [
+                    enrollments[str(i)]
+                    for i in assignment.assignees
+                    if str(i) in enrollments
+                ]
+                a["assignee_names"] = [
+                    self._full_name(self._find_student(students, a)) for a in assignees
+                ]
```
stephenfuqua, line 202: Typically an assignment is given to an entire class, but in Schoology you can also give it to individuals. This flattens the list into a single column.

experimental/schoology/export-assignment-data.py
```diff
@@ -0,0 +1,269 @@
...
+                "id": section.id,
+                "course_id": section.course_id,
+                "school_id": section.school_id,
+                "course_title": section.course_title,
+                "course_code": section.course_code,
+                "section_title": section.section_title,
+                "grading_period_ids": section.grading_periods,
+                "grading_period_names": [
+                    self.grading_periods_lookup[gp] for gp in section.grading_periods
+                ],
```
stephenfuqua, line 152: A section can be in multiple grading periods, for example ["first six weeks", "fall semester", "2020-2021 school year"]. In Slack we discussed creating separate output records for each school year. I ran out of time to change my code; for now the list is flattened into a single column.

experimental/schoology/export-assignment-data.py
```diff
@@ -0,0 +1,269 @@
...
+            }
+
+        return [dto_to_section(s) for s in section_list]
+
+    def _full_name(self, student):
+        names = [student["name_first"], student["name_middle"], student["name_last"]]
+        return " ".join(x for x in names if x)
+
+    def _find_student(self, students, uid):
+        return next(s for s in students if s["uid"] == uid)
```
stephenfuqua, line 166: If you haven't seen this yet, it is like the C# `.FirstOrDefault(x => x["uid"] == uid)`. Arguably this should protect against no match found by having a try / catch on `StopIterator`. In theory it should never happen, but probably want to handle gracefully on the 1 in a million(???) chance it _does_ occur.

experimental/schoology/export-assignment-data.py
```diff
@@ -0,0 +1,269 @@
...
+
+        return [dto_to_student(s) for s in student_list]
+
+    def _get_sections(self):
+        # TODO: planning to build a fresh Schoology API package, at which time we
+        # can delete the following extension method
+        from schoolopy.models import Section
+
+        def get_sections_for_course(Self):
+            resource = "courses/{}/sections".format(Self.id)
```
stephenfuqua, line 118: Opportunity to use f-string instead.

experimental/schoology/export-assignment-data.py
```diff
@@ -0,0 +1,269 @@
...
+                "section_id": assignment["section_id"],
+                "section_title": assignment["section_title"],
+                "assignment_id": assignment["id"],
+                "assignment_title": assignment["title"],
+                "revision_id": submission.revision_id,
+                "student_uid": submission.uid,
+                "student_name": self._full_name(
+                    self._find_student(students, str(submission.uid))
+                ),
+                "created": datetime.fromtimestamp(submission.created),
```
stephenfuqua, line 234: This gives us a nice human readable format instead of a Unix timestamp.

experimental/schoology/export-assignment-data.py
```diff
@@ -0,0 +1,269 @@
...
+        self.sc_client = schoolopy.Schoology(
+            schoolopy.Auth(self.api_key, self.api_secret)
+        )
+
+    def _get_grading_periods(self):
+        gp_list = self.sc_client.get_grading_periods()
+        gl = {
+            gp.id: gp.title for gp in gp_list if gp.id in self.desired_grading_periods
+        }
+        self.grading_periods_lookup = gl
```
stephenfuqua, line 80: We only want data "for the current school year." There is no native concept of that in Schoology. So there are grading_period_ids in the environment variables to identify the periods of interest. I intended to include a small script to help someone look up the correct IDs but didn't get around to it yet. This function looks up all of the grading periods in the API and then filters to have only the desired ones, storing them in a lookup table for reference later on.

bradbanister (approved):
Looks good to me.

lalrg (approved):

PR #28, opened 2020-09-17
[Fizz-61] Renormalize Google Classroom

bradbanister:


lalrg (commented):
The code looks great now.

src/google/google-extractor/api/api_caller.py
```diff
@@ -0,0 +1,45 @@
...
+        is the list of dicts of the API response accumulated across pages
+
+    Returns
+    -------
+    list
+        a list of dicts of the API response property requested
+    """
+    if results is None:
+        results = []
+    response = execute(resource_method(**resource_parameters))
```
stephenfuqua, line 37: How are 4xx and 5xx responses handled?

src/google/google-extractor/api/courses.py
```diff
@@ -0,0 +1,19 @@
...
+
+def request_courses(resource):
+    return call_api(
+        resource.courses().list,
+        {"courseStates": ["ACTIVE"]},
+        "courses",
+    )
+
+
+def courses_dataframe(resource):
```
stephenfuqua, line 13: It is trivial enough to read the code and see the difference between these. If using auto-complete, it would be harder to tell what needs to occur. Aside from the question of docstring - which is nice to have but I don't see it as mandatory - I wonder about changing the function name to include a verb, and to help the caller know that this is actually making an API request?

stephenfuqua (changes requested):
Just requesting an update to the function naming pattern.

src/google/google-extractor/api/courses.py
bradbanister: changing this naming pattern to "request_xyz_as_df"

src/google/google-extractor/api/api_caller.py
bradbanister: in this code, 4xx and 5xx are blanket retried with exponential backoff with a 60 second max.  then it crashes with an uncaught exception (oops).  will update to instead log the exception and move on.  future code will be finer grained e.g. handle 429 rate limiting, 401 unauthorized, etc.

stephenfuqua: Uncaught exception might be good enough for now. I believe we have ticket for adding logging to the application later.

stephenfuqua (approved):

PR #31, opened 2020-09-18
[FIZZ-83] Sample output files

stephenfuqua:
I couldn't figure out what the `LMSGrade` model represents, as distinct from the `Grade` field on the `AssignmentSubmission` model, so there is no sample of this model.


bradbanister (approved):
LGTM

stephenfuqua:
> I couldn't figure out what the `LMSGrade` model represents, as distinct from the `Grade` field on the `AssignmentSubmission` model, so there is no sample of this model.

@ggaronzik explained after standup that some of these systems allow a grade to be recorded directly on the enrollment, rather than only on the assignment.

stephenfuqua:
@ggaronzik can you take one more look  now that I've made modifications based on your review comments in Jira?

stephenfuqua:
All review comments have been addressed. I'm going to merge this. We can always modify again later if we find required tweaks.

PR #34, opened 2020-09-22
[FIZZ-80] Load User File

stephenfuqua:
Early draft, establishing the command line user interface with complete set of tests.

Other than the password, I'm not sure that it makes sense to set any of these values with environment variables. Thoughts?


src/lms-ds-installer/mssql/0001-table-user.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE lms.[User] (
```
stephenfuqua, line 1: This will eventually be replaced by output from MetaEd. Need something work with until then.

bradbanister:
I like the idea of trying out [ConfigArgParse](https://pypi.org/project/ConfigArgParse/).  Drop in replacement for argparse that adds automatic environment variable and config file support in addition to command line parameters.

stephenfuqua:
> I like the idea of trying out [ConfigArgParse](https://pypi.org/project/ConfigArgParse/). Drop in replacement for argparse that adds automatic environment variable and config file support in addition to command line parameters.

Definitely looks interesting. Not one of the options I came across in my limited searching yesterday. Most recent release was April - somewhat active project. Large number of issues. But I'm guessing a lot of these are edge cases. It is worth checking out.

src/lms-ds-loader/lms_ds_loader/arguments.py
```diff
@@ -0,0 +1,47 @@
...
+
+        if not port:
+            port = 1433
+        if type(port) == str and port.strip() == "":
+            port = 1433
+
+        return f"mssql+pyodbc://{username}:{password}@{server},{port}/{db_name}?driver=SQL Server"
+
+
+class Arguments:
```
bradbanister, line 39: For these kind of simple classes, check out [dataclass](https://docs.python.org/3/library/dataclasses.html), new in Python 3.7. It's an easier way to define them and provides conversions to Dicts and Tuples if you so choose.

src/lms-ds-loader/lms_ds_loader/arguments.py
```diff
@@ -0,0 +1,47 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+
+class DbConnection:
+    @staticmethod
+    def build_for_mssql_with_integrated_security(server, port, db_name):
```
bradbanister, line 9: if DbConnection isn't going to hold any state, consider moving these static methods to a DbConnection module as plain functions.

src/lms-ds-installer/pyproject.toml
```diff
@@ -0,0 +1,15 @@
+[tool.poetry]
+name = "lms-ds-installer"
+version = "0.1.0"
+description = "Learing Management System Data Store (LMS-DS) Installer"
+authors = ["Ed-Fi Alliance, LLC and Contributors"]
+
```
lalrg, line 6:
```suggestion
license = "Apache-2.0"

```

src/lms-ds-loader/.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+max-line-length = 110
```
lalrg, line 2: We are using 88 for other projects... We should define a general line length.

src/lms-ds-loader/.pylintrc
```diff
@@ -0,0 +1,14 @@
...
+ignore = .venv
+
+[MESSAGES CONTROL]
+disable =
+    C0330,
+    C0326,
+    fixme,
+    line-too-long,
+    missing-function-docstring,
+    missing-module-docstring
```
lalrg, line 11: We could create tickets for adding docstrings and remove the related "disable" lines in all projects.

src/lms-ds-loader/.pylintrc
```diff
@@ -0,0 +1,14 @@
...
+
+[format]
+max-line-length = 88
```
lalrg, line 14: .flake8 says 110 and .pylint says 88

src/lms-ds-loader/lms_ds_loader/arguments.py
lalrg: Yeah, the main reason for adding classes would be to encapsulate some data, if you won't, you could use simple functions.

src/lms-ds-loader/lms_ds_loader/arguments.py
```diff
@@ -0,0 +1,47 @@
...
+        if type(port) == str and port.strip() == "":
+            port = 1433
+
+        return f"mssql+pyodbc://{username}:{password}@{server},{port}/{db_name}?driver=SQL Server"
+
+
+class Arguments:
+    csv_path = ""
+    engine = ""
+    db_connection = None
```
lalrg, line 42: Are you aware that this variables don't referenciate the instance ones? They will act similar to "static" values from the class, while the ones defined in the __init__ method, will act as class instance's attributes.

src/lms-ds-loader/tests/test_lms_ds_loader.py
```diff
@@ -0,0 +1,5 @@
+from lms_ds_loader import __version__
+
+
+def test_version():
+    assert __version__ == '0.1.0'
```
lalrg, line 5: Should we keep this kind of auto-generated tests?

src/lms-ds-loader/.pylintrc
bradbanister: black defaults to 88

stephenfuqua: I've changed both to 110. There are some lines that Black can't reformat: those with long strings. I don't want flake/pylint to complain about those longer lines, up to 110 characters.

src/lms-ds-loader/.pylintrc
stephenfuqua: Actually, I was intending to add docstrings everywhere before completing the ticket. For me that's part of the definition of done. I just copied this file from somewhere else and didn't inspect it closely enough. I'm going to remove lines 10 and 11..

src/lms-ds-loader/.flake8
stephenfuqua: 110 is the accepted standard in the Ed-Fi coding conventions.

src/lms-ds-installer/pyproject.toml
stephenfuqua: You can do that in a pyproject.toml file? Nice.

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: Good call. Originally this did encapsulate data, but evolved into just helper methods. Static class of helper methods can be very useful, but not necessary here.

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: Will take a look.

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: Done.

src/lms-ds-installer/pyproject.toml
stephenfuqua: Done

src/lms-ds-loader/.pylintrc
stephenfuqua: Removed those. Will add docstrings before the PR is merged..

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: Hm, that was not the impression I had, but I probably wasn't reading carefully. I definitely want instance methods here. Will look into this further.

src/lms-ds-loader/tests/test_lms_ds_loader.py
stephenfuqua: I haven't decided what to do with this yet. If I don't add any real tests here then I'll remove the file.

src/lms-ds-loader/.pylintrc
stephenfuqua: Docstrings have been added.

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: I removed these, only setting them in the constructor

src/lms-ds-loader/tests/test_lms_ds_loader.py
stephenfuqua: Removed this file.

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: I've deferred looking into that, can refactor in the next ticket if desired.

src/lms-ds-loader/tests/test_csv_to_sql.py
```diff
@@ -0,0 +1,49 @@
...
+            To see that message, you must call pytest with argument `-s`. """
+            # print(to_sql_mock.call_args_list)
+            to_sql_mock.assert_called_with(
+                table,
+                sal_mock.return_value.connect.return_value.__enter__.return_value,
+                schema="lms",
+                if_exists="append",
+                index=False,
+                method="multi",
+            )
```
stephenfuqua, line 49: This was the most difficult mock.... made more difficult originally because I accidentally called `create_engine.connect()` instead of `create_engine().connect()`!

src/lms-ds-loader/tests/test_file_processor.py
```diff
@@ -0,0 +1,67 @@
...
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import pytest
+
+from lms_ds_loader.file_processor import FileProcessor
+from lms_ds_loader.csv_to_sql import CsvToSql
+
+
+def create_stub(name, data):
+    return type(name, (object,), data)
```
stephenfuqua, line 13: This little helper method creates very simple stub objects.

```
file_system = create_stub(
    "LmsFileystemProcessor", {"Users": [file_1, file_2]}
)

file_system.Users == [file_1, file_2]
```

src/lms-ds-loader/tests/test_lms_filesystem_provider.py
```diff
@@ -0,0 +1,92 @@
...
+            with pytest.raises(OSError):
+                LmsFilesystemProvider(csv_path).get_all_files()
+
+        def test_given_valid_path_does_not_contain_user_directory_then_do_nothing(
+            self, fs
+        ):
+            csv_path = "exists"
+
+            # Arrange
+            fs.create_dir(csv_path)
```
stephenfuqua, line 37: This is using [pyfakefs](https://pypi.org/project/pyfakefs/) to fully mock out the filesystem.

src/lms-ds-loader/tests/test_csv_to_sql.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class Test_when_reading_csv_and_loading_into_a_database:
+    class Test_given_valid_arguments:
+        def test_then_load_file_into_specified_table_using_pandas(self, mocker):
+            connection_string = "23456789"
+            file = "a/b/c.csv"
+            table = "c"
+
+            # Arrange
```
stephenfuqua, line 20: I often put these `# Arrange`, `# Act`, `# Assert` comments into tests to help the reader better understand what is going one. It stems from my frustration at looking at long tests and not even being able to find the "action" amidst all of the noise - the the thing being tested.

src/lms-ds-loader/.coveragerc
```diff
@@ -0,0 +1,6 @@
+[run]
```
lalrg, line 1: Cool! I did not know we could add this file.

src/lms-ds-loader/.pylintrc
```diff
@@ -0,0 +1,12 @@
+[MASTER]
+ignore = .venv
+
+[MESSAGES CONTROL]
+disable =
+    C0330,
+    C0326,
+    fixme,
+    line-too-long
```
lalrg, line 9: Why is line-too-long disabled?

src/lms-ds-loader/lms_ds_loader/file_processor.py
```diff
@@ -0,0 +1,40 @@
...
+
+class FileProcessor:
+    """
+    Processes all files in the LMS-UDM compatible file system and calls a
+    program to read and load them into a database.
+
+    Parameters
+    ----------
+    file_system : LmsFilesystemProvider
+        A fully-prepared instance of `LmsFilesystemProvider`.
```
stephenfuqua, line 18: I thought about combining `FileProcessor` and `LmsFilesystemProvider` into one class. Single Responsibility Principle encourages me to keep two classes. 

As we look at building analyzers, I'm thinking of moving `LmsFilesystemProvider` to a stand-alone package so that it can be used by those analyzers. `FileProcessor` might become a strategy class that is injected into `LmsFileystemProvider`, or it might stay the way it is right now. No reason to decide that right now.

src/lms-ds-loader/lms_ds_loader/constants.py
```diff
@@ -0,0 +1,15 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+
```
lalrg, line 6: The Idea of this file is to have a kind of namespaces, right?

src/lms-ds-loader/.pylintrc
stephenfuqua: I copied this from Google Classroom Extractor. @bradbanister ?

src/lms-ds-loader/lms_ds_loader/constants.py
stephenfuqua: Yes. I wanted to re-use these constants - and there will be more eventually - in a few places. I like grouping constants by similarity so that it is easy to find them. Hence "namespace-like" use of nested classes.

src/lms-ds-loader/tests/test_argparser.py
```diff
@@ -0,0 +1,352 @@
...
+    USERNAME = "my_user"
+    PASSWORD = "my_password"
+    PORT = 1234
+
+    def path_args(self):
+        return ["--csvpath", self.PATH]
+
+    def engine_args(self, engine):
+        return ["--engine", engine]
+
```
lalrg, line 26: This could work for the methods that don't require parameters. So when you call them you can do it like:
`object.server_args`

```suggestion

@property
```

src/lms-ds-loader/tests/test_csv_to_sql.py
lalrg: Looks good

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: I like what I see with the `dataclass`. But there is something confusing about it. @lalrg pointed out that `my_var` in the sample below is a _class variable_, not an _instance variable_ as I had assumed ([confirmation](https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables)).

```python
class MyClass:
    my_var: str = None
```

Now if we write that as a data class, `my_var` is now an instance variable. So if you don't read the class carefully - and miss that it is a `dataclass`, then this can be a little surprising. FYI, [here's a test](https://ideone.com/dnwfo5) that proves it is an instance var, not class var.

```python
@dataclass
class MyClass:
    my_var: str
```

Side note: there is a really import point on mutable instance variables (list, dict, tuple) in the [dataclass](https://docs.python.org/3/library/dataclasses.html#mutable-default-values).

src/lms-ds-loader/lms_ds_loader/argparser.py
```diff
@@ -0,0 +1,77 @@
...
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import configargparse
+from lms_ds_loader.constants import Constants
+from lms_ds_loader.arguments import Arguments
+
+
+def parse_arguments(args_in) -> Arguments:
```
stephenfuqua, line 11: What do you guys think - better to have this bare function here, or move this to the Arguments class as a static method?

lalrg: I like the current approach don't see a lot of value on adding this to the Arguments class.

src/lms-ds-loader/lms_ds_loader/arguments.py
```diff
@@ -0,0 +1,97 @@
...
+        Base path for finding CSV files.
+    engine : str
+        Database engine, either "mssql" or "postgresql"
+    """
+
+    def __init__(self, csv_path, engine):
+        assert csv_path is not None, "csv_path cannot be None"
+        assert csv_path.strip() != "", "csv_path cannot be whitespace"
+        assert engine is not None, "engine cannot be None"
+        assert engine.strip() != "", "engine cannot be whitespace"
```
stephenfuqua, line 25: If I convert this to a `dataclass`, then I'll lose my guard clauses, since __init__ will be generated automatically. There's a case to be made that I'm overdoing it with these guard clauses, but it will require someone swimming hard upstream to convince me of that case. I learned the hard way not to trust input data.

If I lose the guard clauses here, then I'll just have to guard against them elsewhere. But, since these values are not `readonly`, that might even be a better idea. After all, someone can always do this:

```
a = Arguments("some/path", "mssql")
a.engine = None

a.some_method()
```

And now if `some_method` were relying on `engine` not being `None`, then there might be an unexpected error. So `some_method` should really be checking for valid state of vars before using them. In which case checking them in the constructor was unnecessary.

I might have just talked myself into using the `@dataclass` attribute and improving the error handling closer to where it actually matters. I'll do that in the next story, the UPDATEs story.

src/lms-ds-loader/lms_ds_loader/argparser.py
stephenfuqua: This seems like a Pythonic approach. Just feels a little weird having this function without a class 😄 . In C# I definitely would not have a separate class for this function - it would be a builder method on the `Arguments` class. But in Python we have this flexibility. I'm not sure which model is better for future maintenance, really they feel about equal to me. I do not yet have a strong feeling, even though it feels a bit strange leaving this here.

src/lms-ds-installer/README.md
```diff
@@ -0,0 +1,4 @@
+# LMS-DS-Installer
+
+Scripts for installing the Ed-Fi Learning Management System Data Store (LMS-DS)
+sql schema script. Further details to follow when ready.
```
lalrg, line 4: Will you leave it like this for now?

lalrg: Would be helpful to redirect the users to the readme of src/lms-ds-loader when they get here

src/lms-ds-loader/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+   ```bash
+   poetry install
+   ```
+
+1. Run the two SQL scripts in `src/lms-ds-installer/mssql` in a database.
+
+## Running the Tool
+
+For detailed help, execute `poetry run lms_ds_loader/main.py -h`.
```
lalrg, line 26:
```suggestion
For detailed help, execute `poetry run python.exe .\lms_ds_loader\main.py -h`.

```

lalrg (approved):
Just a couple small changes in readme files, you can update it and merge.

src/lms-ds-installer/README.md
stephenfuqua: I was going to leave like this for now, until [FIZZ-88](https://tracker.ed-fi.org/browse/FIZZ-88), which is the ticket for working on this installer. I can add a little more information right now though, for anyone who happens to pick up this source code.

src/lms-ds-loader/README.md
stephenfuqua: done

PR #35, opened 2020-09-23
[FIZZ-44] Create FIZZ MetaEd Project and Downstream Artifacts

ggaronzik:


docs/metaed-project/Association/UserLMSSectionAssociation.metaed
```diff
@@ -0,0 +1,32 @@
+Association UserLMSSectionAssociation
+    documentation ""
+    domain entity User
+        documentation ""
+    domain entity LMSSection
+        documentation ""
+    shared integer Identifier
```
stephenfuqua, line 7: This is the place holder for the serial Id? I'm not opposed to `Identifier` fully-spelled out - just curious why spell it out instead of `Id`?

stephenfuqua:
@ggaronzik please rebase from main

Is it necessary to put the MetaEdOutput folder into source control? I always get squeamish when I see large files (for example, > 1 MB) in source control. If this file can easily be regenerated from MetaEd, I'm wondering if we can exclude it from source control. If that is the case, then we will need to remove the directory and rewrite the existing commits to completely strip it out of the filesystem - before a merge occurs.

docs/metaed-project/Association/UserLMSSectionAssociation.metaed
ggaronzik: Yes. I updated these to be the full "Identifier" from "Id" to be consistent with the other identifier attributes. If it makes more sense to shorten these to just Id, I'm good with that modification.

ggaronzik:
> Is it necessary to put the MetaEdOutput folder into source control? I always get squeamish when I see large files (for example, > 1 MB) in source control. If this file can easily be regenerated from MetaEd, I'm wondering if we can exclude it from source control. If that is the case, then we will need to remove the directory and rewrite the existing commits to completely strip it out of the filesystem - before a merge occurs.

Not necessary. If we pull the MetaEdOutput folder I do think we should identify artifacts (like the database scripts and possibly schema) to keep in source control and where those should live.

docs/metaed-project/Association/UserLMSSectionAssociation.metaed
stephenfuqua: `Id` is more typical and takes less typing. Part of me actually likes the written out form better. You would need to change the UDM to `Id` (although that might be a simple find-and-replace operation in the drawio XML contents). Lets keep `Identifier` since that's what we agreed to with the UDM.

stephenfuqua (changes requested):
I'm locking this to prevent merge.  Replaced with https://github.com/Ed-Fi-Exchange-OSS/Ed-Fi-X-Fizz/pull/37. Once PR 37 is accepted, I will need to do a little more clean up on the FIZZ-44 branch before closing it, stripping out the last two commits. I think that will force remove them from the Git database.

PR #36, opened 2020-09-23
FIZZ-46 Rewrite Schoology extractor

lalrg:


.vscode/settings.json
```diff
@@ -74,7 +74,13 @@
...
         "autoHeader": "autoSave",
         "exclude": [
-            "css", "json", "xml", "config", "plaintext", "markdown", "batch"
+            "css",
+            "json",
+            "xml",
+            "config",
+            "plaintext",
+            "markdown",
+            "batch"
```
stephenfuqua, line 83: I'm surprised to see this file change, and worried that someone else's changes will reformat again. I can accept this for now, but let's take care to watch out for frequent format changes - if we keep going back and forth with these little updates then we need to standardize something, somewhere, so that we stop getting changes.

lalrg: Agreed

src/schoology-extractor/data_extractor/lib/assignments.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Attributes:
+        request_client (RequestClient): The client used for authenticate and send requests to Schoology
+
+    """
+
+    def __init__(self, SCHOOLOGY_KEY, SCHOOLOGY_SECRET):
+        self.request_client = RequestClient(SCHOOLOGY_KEY, SCHOOLOGY_SECRET)
+
+    def get_by_section_id_array(self, section_id_array):
```
stephenfuqua, line 20: please add guard assertion to ensure that `section_id_array` is not `None`. Might also make sense to ensure that it is a list. Finally, might consider removing the word "array" from the argument name. Maybe just `section_ids`?

src/schoology-extractor/data_extractor/lib/export_data.py
```diff
@@ -0,0 +1,9 @@
+import pandas as pd
+
+
+def tocsv(data, output_path):
```
stephenfuqua, line 4: Please add guard assertions to make sure that neither argument is `None`.

stephenfuqua: And think about where appropriate in other functions below. I'm not going to comment on every one of them.

src/schoology-extractor/data_extractor/lib/users.py
```diff
@@ -0,0 +1,22 @@
...
+        request_client (RequestClient): The client used for authenticate and send requests to Schoology
+
+    """
+
+    def __init__(self, SCHOOLOGY_KEY, SCHOOLOGY_SECRET):
+        self.request_client = RequestClient(SCHOOLOGY_KEY, SCHOOLOGY_SECRET)
+
+    def get_all(self):
+        response = self.request_client.get("users")
+        return response["user"]
```
stephenfuqua, line 22: What happens if the dictionary does not contain an element `users`?

src/schoology-extractor/data_extractor/lib/sections.py
```diff
@@ -0,0 +1,22 @@
...
+        request_client (RequestClient): The client used for authenticate and send requests to Schoology
+
+    """
+
+    def __init__(self, SCHOOLOGY_KEY, SCHOOLOGY_SECRET):
+        self.request_client = RequestClient(SCHOOLOGY_KEY, SCHOOLOGY_SECRET)
+
+    def get_by_id(self, section_id):
+        response = self.request_client.get(f"sections/{section_id}")
+        return response
```
stephenfuqua, line 22: The various classes like `Sections`, `Submissions`, `Users`, etc. are very small. What do you think about moving `get_by_id` into the `RequestClient` and changing the name to `get_sections`? Then you wouldn't need to instantiate so many objects, turning `RequestClient` into a fatter façade for making calls to the Schoology API.

src/schoology-extractor/data_extractor/main.py
```diff
@@ -0,0 +1,50 @@
+import os
+from lib import export_data
+from lib.users import Users
+from lib.assignments import Assignments
+from lib.sections import Sections
+from lib.submissions import Submissions
+from dotenv import load_dotenv
```
stephenfuqua, line 7: I _think_ this is more like the usual standard: standard libraries, 3rd party, and then custom imports.

```suggestion
import os

from dotenv import load_dotenv

from lib import export_data
from lib.users import Users
from lib.assignments import Assignments
from lib.sections import Sections
from lib.submissions import Submissions
```

src/schoology-extractor/tests/test_request_client.py
```diff
@@ -0,0 +1,59 @@
...
+    class TestInitMethod:
+        def test_default_base_url_is_set(self):
+            request_client = RequestClient(pytest.FAKE_KEY, pytest.FAKE_SECRET)
+            assert request_client.base_url == pytest.DEFAULT_URL
+
+        def test_custom_base_url_is_set(self):
+            custom_url = "a_custom_url"
+            request_client = RequestClient(
+                pytest.FAKE_KEY, pytest.FAKE_SECRET, custom_url)
+            assert request_client.base_url == custom_url
```
stephenfuqua, line 25: I like to use a naming convention that is inpsired by Behavior Driven Development, using the words "given, when, then". As you have done here, I usually have nested classes. Typically I switch to "when given then": when structured that way, the action (given) is at a top level, with "when... then" at a more detailed level to encapsulate the variations. What do you think about this?
```suggestion
class TestRequestClient:
    class Test_when_constructing:
        def test_given_not_passing_in_a_url_then_use_the_default_url(self):
            request_client = RequestClient(pytest.FAKE_KEY, pytest.FAKE_SECRET)
            assert request_client.base_url == pytest.DEFAULT_URL

        def test_given_custom_url_then_use_it_for_the_base_url(self):
            custom_url = "a_custom_url"
            request_client = RequestClient(
                pytest.FAKE_KEY, pytest.FAKE_SECRET, custom_url)
            assert request_client.base_url == custom_url
```

src/schoology-extractor/tests/test_request_client.py
```diff
@@ -0,0 +1,59 @@
...
+        def test_oauth_present_in_header(self, default_request_client):
+            assert self._present_in_header(
+                default_request_client._request_header,
+                header_key="Authorization",
+                contained_value="OAuth"
+            )
+
+        def _present_in_header(self, request_header, header_key, contained_value):
+            header_section = request_header[header_key]
+            return header_section.find(contained_value) != -1
```
stephenfuqua, line 44: I like this method!

src/schoology-extractor/.gitignore
```diff
@@ -0,0 +1 @@
+!lib
```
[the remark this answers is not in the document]
lalrg: Yes

src/schoology-extractor/tests/test_request_client.py
lalrg: I will change it

src/schoology-extractor/.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "python.linting.pylintEnabled": true,
+  "python.pythonPath": ".venv\\Scripts\\python.exe"
+}
```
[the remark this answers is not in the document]
lalrg: I committed this file by error...  Do you still want to keep it?

src/schoology-extractor/.pylintrc
```diff
@@ -0,0 +1,11 @@
...
+disable = 
+    C0330,
+    C0326,
+    fixme,
+    line-too-long,
+    missing-function-docstring,
+    missing-module-docstring
+
+[format]
+max-line-length = 88
```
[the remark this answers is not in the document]
lalrg: I'll change it

src/schoology-extractor/data_extractor/lib/assignments.py
```diff
@@ -0,0 +1,28 @@
+from .request_client import RequestClient
```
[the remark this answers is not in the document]
lalrg: Ok, thanks

src/schoology-extractor/data_extractor/lib/assignments.py
```diff
@@ -0,0 +1,28 @@
...
+    The Assignments class contains methods to get assignments' information from Schoology API
+
+    Args:
+        SCHOOLOGY_KEY (str): The consumer key given by Schoology
+        SCHOOLOGY_SECRET (str): The consumer secret given by Schoology
+
+    Attributes:
+        request_client (RequestClient): The client used for authenticate and send requests to Schoology
+
+    """
```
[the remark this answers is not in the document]
lalrg: I don't have any strong preference. The only reason why I have used google format is because it seemed more readable to me.

src/schoology-extractor/data_extractor/lib/assignments.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, SCHOOLOGY_KEY, SCHOOLOGY_SECRET):
+        self.request_client = RequestClient(SCHOOLOGY_KEY, SCHOOLOGY_SECRET)
```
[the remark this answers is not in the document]
lalrg: I will add the asserts and change to lowercase.

src/schoology-extractor/.vscode/settings.json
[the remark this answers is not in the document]
stephenfuqua: If setting the python path there is helpful, then keep the file.

src/schoology-extractor/data_extractor/lib/request_client.py
```diff
@@ -0,0 +1,55 @@
...
+        auth_header += 'oauth_nonce="%s",' % "".join(
+            [str(random.randint(0, 9)) for i in range(8)]
+        )
+        auth_header += 'oauth_timestamp="%d",' % time.time()
+        auth_header += 'oauth_signature_method="PLAINTEXT",'
+        auth_header += 'oauth_version="1.0",'
+        auth_header += 'oauth_signature="%s%%26%s"' % (
+            self.consumer_secret,
+            "",
+        )
```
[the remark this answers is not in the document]
lalrg: To me it does not make a big difference in readability. Both ways are ugly because of they contain a lot of hardcoded text, and that's the reason why it is in the _request_header property. But I can change it if you want, is a small change.

src/schoology-extractor/data_extractor/lib/assignments.py
lalrg: I will make the change

src/schoology-extractor/data_extractor/lib/users.py
lalrg: It could fail, I'll fix it.

src/schoology-extractor/data_extractor/lib/sections.py
lalrg: And what do you think about keeping classes but only as properties in RequestClient? I am thinking about keeping files small, if we wrap them inside of RequestClient while keeping the classes for separating the code, we can avoid creating too many instances in the main.py by moving them to the __init__ method of the RequestClient, but we'll keep the possibility of adding more code to the classes without making the RequestClient a single big class.

src/schoology-extractor/data_extractor/main.py
lalrg: I will change it

src/schoology-extractor/.vscode/settings.json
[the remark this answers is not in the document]
lalrg: It could be helpful but only if you open vsCode with the src/schoology-extractor as the root folder. I think I'll keep it with the required modifications.

src/schoology-extractor/data_extractor/lib/sections.py
stephenfuqua: I don't have a _strong_ feeling about this. I've normally used, and normally seen, a single client with lots of methods for getting what you need from a remote server - whether that is an API or a database. Thinking about a database repository class that works on the entire database, or the Entity Framework DbContext class. It is true that I've also seen one repository per table, but it just seems like overkill and more maintenance that is needed.

If keeping separate classes for resource that will be retrieved, then will they benefit from a base class to reduce code duplication?

lalrg: I don't think they could get a lot of benefit from a base class, I tried that and I ended up adding more lines than the ones I save, and the only thinkg I could share was the initial request client, because there are not many reusable methods, because for example a get_all method, will have a different path structure:
users : /users
assignments: sections/{section_id}/assignments

I think we could  go with a single RequestClient for now, if it gets too big, should not be so difficult to separate some methods in the future.

src/schoology-extractor/data_extractor/lib/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.1.0'
```
stephenfuqua, line 1: Each Python file needs to have the file header on it. Please check that you have the recommended psionic header extension in Visual Studio Code (exact name is in the `.vscode/extensions.json` file). With that extension, every new file you create from within VS Code will automatically get the correct header. A file created outside of VS code will require manual editing. You can pull up the command palette and find the "insert header" command.

Does not apply to configuration files including *.json, ini-type config files, or xml config files.

src/schoology-extractor/.env.example
```diff
@@ -0,0 +1,4 @@
+SCHOOLOGY_KEY=[YOUR_SCHOOLOGY_KEY]
+SCHOOLOGY_SECRET=[YOUR_SCHOOLOGY_SECRET]
+SCHOOLOGY_OUTPUT_PATH=[YOUR_SCHOOLOGY_OUTPUT_PATH]
```
bradbanister, line 3: Since you're providing an output directory location, you may want to remove and just output there. Or at least provide it in the example here.  It took me a few tries to get it right, as "output".

src/schoology-extractor/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.linting.pylintEnabled": true
```
bradbanister, line 2: I'm unclear why this doesn't include `"python.pythonPath": ".venv\\Scripts\\python.exe"`. Are you running a single poetry env for all projects?

src/schoology-extractor/README.md
```diff
@@ -0,0 +1,46 @@
...
+-   `SCHOOLOGY_KEY=<administrator's API key>`
+-   `SCHOOLOGY_SECRET=<administrator's API secret>`
+-   `SCHOOLOGY_OUTPUT_PATH=<output directory>`
+-   `SCHOOLOGY_SECTION_IDS=<csv list of sections' ids>`
+
+## Execution
+
+```bash
+poetry install
+poetry run python.exe main.py
```
bradbanister, line 30: should reflect that main.py is under lib

src/schoology-extractor/data_extractor/main.py
```diff
@@ -0,0 +1,47 @@
+# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: About "data_extractor" directory name, not this file.  Python standard looks like same name for this directory as the project name.  Would suggest "schoology_extractor" since Python doesn't like dashes in import.

src/schoology-extractor/data_extractor/lib/__init__.py
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: About "lib" directory name, not this file.  Since it's a module name, would prefer something more descriptive.

src/schoology-extractor/tests/test_lib/__init__.py
```diff
@@ -0,0 +1,4 @@
+# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: About "test_lib" directory name, not this file.  I haven't seen this pattern of prefixing module name being tested with "test_".  I've only seen directory name same as module test name.

src/schoology-extractor/tests/test_lib/test_request_client.py
```diff
@@ -0,0 +1,231 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import pytest
+
+from data_extractor.lib.request_client import RequestClient
+
+pytest.fake_key = 'TEST_KEY'
```
bradbanister, line 10: Do these have to be added to the pytest module?  I'd think they are fine standalone.
Separately, Pylance doesn't like them because they aren't part of the module definition.

src/schoology-extractor/tests/test_lib/test_request_client.py
```diff
@@ -0,0 +1,231 @@
...
+                pytest.fake_key, pytest.fake_secret, custom_url)
+
+            # Assert
+            assert request_client.base_url == custom_url
+
+
+    class Test_when_building_request_header:
+        def test_then_consumer_key_is_present(self, default_request_client):
+            assert self._present_in_header(
+                default_request_client._request_header,
```
bradbanister, line 45: You have default_request_client here hiding the function of the same name in outer scope.  Other places too.

src/schoology-extractor/data_extractor/lib/export_data.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import pandas as pd
+
+
+def tocsv(data, output_path):
```
bradbanister, line 9: consider rename to "to_csv"

src/schoology-extractor/data_extractor/lib/request_client.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+    @property
+    def _request_header(self):
+        """The _request_header property helps to build the Request Header for oauth requests
+            Returns:
+                dict: Request headers
+        """
+        auth_header = 'OAuth realm="Schoology API",'
+        auth_header += 'oauth_consumer_key="%s",' % self.consumer_key
```
bradbanister, line 42: Consider f-strings. Also consider multi-line strings: https://stackoverflow.com/questions/10660435/pythonic-way-to-create-a-long-multi-line-string

bradbanister (changes requested):

src/schoology-extractor/.env.example
lalrg: I will modify the example.

src/schoology-extractor/.vscode/settings.json
lalrg: I did it by mistake, I'll replace line two with the pythonPath

src/schoology-extractor/data_extractor/lib/__init__.py
lalrg: done

src/schoology-extractor/tests/test_lib/__init__.py
lalrg: Fixed

src/schoology-extractor/tests/test_lib/test_request_client.py
lalrg: You are right, I removed the pytest part and it worked.

src/schoology-extractor/tests/test_lib/test_request_client.py
lalrg: Actually, it is a test fixture, by having the same name, I am kind of "injecting" the result of the default_request_client() function.

More information [here](https://docs.pytest.org/en/stable/fixture.html#fixtures-as-function-arguments)

src/schoology-extractor/data_extractor/lib/export_data.py
lalrg: Done

bradbanister (approved):
looks good.  i also ran it against our sandbox and it worked fine.

PR #39, opened 2020-09-28
[FIZZ-59] add data synchronization and testing to google classroom extractor

bradbanister:


src/google-classroom-extractor/google_classroom_extractor/api/coursework.py
```diff
@@ -0,0 +1,44 @@
...
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import logging
+from typing import List, Dict
+import pandas as pd
+from .api_caller import call_api
+
+
+def request_coursework(resource, course_id: str) -> List[Dict[str, str]]:
```
lalrg, line 12: Types annotations look great.

src/google-classroom-extractor/google_classroom_extractor/api/students.py
```diff
@@ -0,0 +1,45 @@
...
+def request_latest_students_as_df(resource, course_ids: List[str]) -> pd.DataFrame:
+    logging.info("Pulling student data")
+    students: List[Dict[str, str]] = []
+    for course_id in course_ids:
+        students.extend(request_students(resource, course_id))
+
+    return pd.json_normalize(students).astype("string")
+
+
+def request_all_students_as_df(resource, course_ids: List[str], sync_db) -> pd.DataFrame:
```
lalrg, line 29: Please consider adding docstrings

src/google-classroom-extractor/google_classroom_extractor/api/usage.py
```diff
@@ -0,0 +1,123 @@
...
+
+def last_sync_date(sync_db) -> Optional[datetime]:
+    with sync_db.connect() as con:
+        try:
+            usage_df = pd.read_sql("SELECT asOfDate FROM Usage", con)
+            if usage_df["asOfDate"].count() == 0:
+                return None
+            return date_parse(usage_df["asOfDate"].max())
+        except OperationalError:
+            # no Usage table yet
```
lalrg, line 36: Could add a logging.error

src/google-classroom-extractor/google_classroom_extractor/main.py
```diff
@@ -0,0 +1,222 @@
...
+from google.oauth2 import service_account
+from sqlalchemy import create_engine
+
+from google_classroom_extractor.api.courses import request_all_courses_as_df
+from google_classroom_extractor.api.coursework import request_all_coursework_as_df
+from google_classroom_extractor.api.students import request_all_students_as_df
+from google_classroom_extractor.api.submissions import request_all_submissions_as_df
+from google_classroom_extractor.api.usage import request_all_usage_as_df
+
+
```
lalrg, line 25: I think it would be nice to simplify a little bit the main.py script, or at least add some comments.

src/google-classroom-extractor/google_classroom_extractor/main.py
```diff
@@ -0,0 +1,222 @@
...
+            students_df=students_df,
+        ),
+        lambda: courses_df,
+        lambda: coursework_df,
+        lambda: submissions_df,
+        lambda: students_df,
+    )
+
+
+if __name__ == "__main__":
```
lalrg, line 205: Instead of executing the same script from both places, I would just use the main for generating the csv and use each notebook to consume the internal packages and include some logic in them, so the notebooks and the main could be used as examples of how to consume the packages.

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+
+
+def request_latest_courses_as_df(resource) -> pd.DataFrame:
+    logging.info("Pulling course data")
+    courses: List[Dict[str, str]] = request_courses(resource)
+    return pd.json_normalize(courses)
+
+
+def request_all_courses_as_df(resource, sync_db) -> pd.DataFrame:
```
lalrg, line 27: Should we add asserts for checking 'None' values in some of the functions?
If yes, we should unit test the none values too.

lalrg (changes requested):
- In the readme file, you should update the instructions for running the tool, `poetry run python google-extractor/main.py` should be `poetry run python google_classroom_extractor/main.py`
- Should the __init__.py files contain the license headers?
- @stephenfuqua and I have added test coverage reports to the code and even when we have not defined a minimum coverage, I think it is a very interesting tool to add to the code and add to the readme, is not required but would be nice and should not take a lot of time.

src/google-classroom-extractor/data/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file - it's the directory placeholder
+!.gitignore
```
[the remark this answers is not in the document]
bradbanister: The alternative is a .gitkeep, but that's by convention.  In addition to placeholding the directory, this blacklists any other file in the directory from inadvertently being checked in.

src/google-classroom-extractor/google_classroom_extractor/api/usage.py
bradbanister: It's actually acceptable behavior when there is no table in db.  Can add a logging.debug().

bradbanister:
Fixed README on running main.py, and added headers to empty __init__.py files

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
bradbanister: Type annotations handle this better, but of course the parameters actually need to be typed.  I'll see if I can find the actual types for resource and sync_db, otherwise asserts are a good idea for untyped values.

src/google-classroom-extractor/google_classroom_extractor/api/students.py
bradbanister: Can do

src/google-classroom-extractor/google_classroom_extractor/main.py
bradbanister: Can refactor

src/google-classroom-extractor/google_classroom_extractor/main.py
bradbanister: Sounds good.  Looking at this now, I should also split pulling data from getting data out of sync db.

bradbanister:
Also fixed notebooks looking in different directory for sync db and bad function name in submissions notebook.

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
stephenfuqua: With or without type annotations, can the system end up with a `None` value on a property before it is dereferenced? If so, I like to guard against that value with an assertion, instead of having an [AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError) raised.

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
```diff
@@ -0,0 +1,40 @@
...
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import logging
+from typing import List, Dict
+import pandas as pd
+from .api_caller import call_api
+
+
```
stephenfuqua, line 11: feels so weird to me not to have a class here. I haven't yet gotten used to that possibility.

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
```diff
@@ -0,0 +1,40 @@
...
+def request_courses(resource) -> List[Dict[str, str]]:
+    # type: ignore
+    return call_api(
+        resource.courses().list,
```
stephenfuqua, line 15: A quick-and-dirty test to confirm something...

```python
class Test_annotations:
    def test_does_type_checking_stop_none(self):

        def my_method_with_str(arg: str):
            print("my_method_with_str: ", str)

        def my_method_with_str_2(arg: str):
            assert arg is not None

        my_method_with_str(None)
        my_method_with_str_2(None)
```

Confirmed: type annotation won't keep you from passing in a `None`. This won't surprise any of us. Just good to see it.

So, what happens if `resource` is `None`? An `AttributeError` would be raised on the attempt to call `courses()`. This is where I would prefer to see the function start with `assert resource is not None`.

src/google-classroom-extractor/tests/courses/test_courses.py
```diff
@@ -0,0 +1,128 @@
...
+def db_room_by_course_id(test_db, course_id) -> int:
+    return test_db.engine.execute(
+        f"SELECT room from Courses WHERE id = '{course_id}'"
+    ).scalar()
+
+
+@patch("google_classroom_extractor.api.courses.request_latest_courses_as_df")
+def test_overlap_removal(mock_latest_courses_df):
+    Path(DB_FILE).unlink(missing_ok=True)
+    test_db = create_engine(f"sqlite:///{DB_FILE}", echo=True)
```
stephenfuqua, line 38: By "formal" definition, this is no longer a unit test - it is connecting to an external resource, and therefore is an integration test. Since sqlite is so easy to use, and we gain value from the integration test, I think I can live with this so long as the user isn't left waiting for the test to complete (which I very much doubt will occur).

src/google-classroom-extractor/tests/courses/test_courses.py
```diff
@@ -0,0 +1,128 @@
...
+    assert db_ending_name(test_db) == "Julia Harding"
+
+    # 2nd pull: courses-2nd-overlaps-1st is 59 rows with 5 overlapping ending with "Maurice Hatfield"
+    mock_latest_courses_df.return_value = pd.read_csv(
+        "tests/courses/courses-2nd-overlaps-1st.csv"
+    )
+    second_courses_df = request_all_courses_as_df(None, test_db)
+    assert dataframe_row_count(second_courses_df) == 59
+    assert db_row_count(test_db) == 71  # 59 new + 17 existing - 5 overlapping
+    assert db_ending_name(test_db) == "Maurice Hatfield"
```
stephenfuqua, line 54: Is there any reasonable way to split this into multiple tests, so that if, for example, line 45 fails, we'll still have a chance to evaluate the block of asserts immediately following line 51?

src/google-classroom-extractor/google_classroom_extractor/main.py
```diff
@@ -0,0 +1,230 @@
...
+    r.normalized_submissions()
+
+    logging.info("Writing data to CSV files")
+    with get_sync_db_engine().connect() as con:
+        for table in con.execute(
+            "SELECT name FROM sqlite_master WHERE type='table';"
+        ).fetchall():
+            table_name = table[0]
+            df = pd.read_sql(f"SELECT * FROM {table_name}", con)
+            df.to_csv(f"{table_name.lower()}.csv", index=False)
```
stephenfuqua, line 230: What happens when a district is using multiple LMS's?

What does the volume of fetched data look like in early December?

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
lalrg: It looks similar to javascript when you just export functions from a file

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
bradbanister: @stephenfuqua @lalrg  I've been thinking about this and it's really quite interesting.  Asserting not null is a useful run-time check in statically-typed languages where null is valid for any object type (like C# before 8.0 introduced the nullable reference type option).  However, in dynamically typed languages like Python, JS, etc. it's actually a rather weak assertion.  Better assertions here would be `assert isinstance(resource, googleapiclient.discovery.Resource)` and `assert isinstance(sync_db, sqlalchemy.engine.base.Engine)`  Python types are not nullable, so these provide a stronger run-time check that it's not just "not None", but actually the type we expect.  Go with this?

bradbanister: FYI, I have a duck-typing example in the Google extractor where all I expect from a "executable_resource" parameter is an object with an "execute" method.  I'm asserting now like so: `assert hasattr(executable_resource, "execute")`

stephenfuqua: You raise a good point. Much better to do `assert type(s) is str` than `assert str is not None`.

lalrg: I have read that most python devs prefer `assert hasattr` over `assert type(s)`

stephenfuqua: It depends on what you're trying to accomplish. Most of the time I'm just trying to see if something is a `str` type. Checking to see if it has a particular string attribute would not prove that it is a string. I don't think I've even see that pattern yet - this thread is the first time I have come across `hasattr` unless I just skimmed past it somewhere.

src/google-classroom-extractor/google_classroom_extractor/api/students.py
bradbanister: Going to add this to FIZZ-77.  Lots of refactoring to happen then.

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
bradbanister: yes, same as JS where a module provides organizational benefit and you don't need state or multiple instances

src/google-classroom-extractor/tests/courses/test_courses.py
bradbanister: they are together because they are dependent.  i'm testing behavior across multiple pulls and asserting correctness after each.

src/google-classroom-extractor/google_classroom_extractor/main.py
bradbanister: I think these are answered by FIZZ-77, Data Output Standardization.  will make sure they are explicit.

src/google-classroom-extractor/google_classroom_extractor/api/submissions.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def request_submissions(
+    resource: Optional[Resource], course_id: str
+) -> List[Dict[str, str]]:
+    assert isinstance(resource, Resource) or resource is None
+    assert isinstance(course_id, str)
+
+    if resource is None:
+        return []
```
lalrg, line 21: Why do you return an empty list instead of just asserting that the resource is not None

lalrg (approved):
Looks good to me.

src/google-classroom-extractor/google_classroom_extractor/api/submissions.py
bradbanister: It's not an assertion.  I'm explicitly allowing None.  That's what the type Optional[Resource] means, Resource or None.
The behavior is then give me no resource, get an empty response.  In practice it's for testability more than anything else, but also a reasonable behavior.

src/google-classroom-extractor/tests/courses/test_courses.py
stephenfuqua: OK

PR #40, opened 2020-09-29
[FIZZ-82] Update existing records based on LastModifiedDate

stephenfuqua:
Refactored the insert record process for faster upload to the database via a staging table, then copy from a staging table to production safely with a script that runs completely db server-side. Then exploit that same system to easily update existing records from the staging table.

### DIFF
--- a/src/lms-ds-loader/README.md
+++ b/src/lms-ds-loader/README.md
@@ -37,6 +37,10 @@ poetry run python ./lms_ds_loader/main.py --server localhost --dbname fizz --use
 1. Run unit tests: `poetry run pytest`
 1. Run unit tests with XLM report: `poetry run pytest --junitxml="pytest.xml"`
 1. Run unit tests with code coverage: `poetry run coverage run -m pytest`
-1. View coverage report in the console: `poetry run coverage report`
-1. Generate an XML coverage report: `poetry run coverage xml` generates file
-   `coverage.xml`
+1. View code coverage:
+    * View coverage report in the console: `poetry run coverage report`
+    * Generate an XML coverage report: `poetry run coverage xml` generates file
+      `coverage.xml`
+    * Generate an HTML file to view in a browser: `poetry run coverage html`
+      generates file  `htmlcov/index.html`. *This version is clickable*,
+      allowing drilldown to see coverage details.
--- a/src/lms-ds-loader/README.md
+++ b/src/lms-ds-loader/README.md
@@ -34,6 +34,9 @@ poetry run python ./lms_ds_loader/main.py --server localhost --dbname fizz --use
 
 ## Dev Operations
 
+Assuming `poetry install` has already been run:
+
+1. Style check: `poetry run flake8`
 1. Run unit tests: `poetry run pytest`
 1. Run unit tests with XLM report: `poetry run pytest --junitxml="pytest.xml"`
 1. Run unit tests with code coverage: `poetry run coverage run -m pytest`
--- a/src/lms-ds-loader/lms_ds_loader/argparser.py
+++ b/src/lms-ds-loader/lms_ds_loader/argparser.py
@@ -23,7 +23,7 @@ def parse_arguments(args_in) -> Arguments:
         A populated `Arguments` object.
     """
 
-    assert type(args_in) is list, "Argument `args_in` must be a list"
+    assert isinstance(args_in, list), "Argument `args_in` must be a list"
 
     parser = configargparse.ArgParser()
     parser.add("-c", "--csvpath", help="base path for input files", required=True)
--- a/src/lms-ds-loader/lms_ds_loader/argparser.py
+++ b/src/lms-ds-loader/lms_ds_loader/argparser.py
@@ -8,6 +8,7 @@ from lms_ds_loader.constants import Constants
 from lms_ds_loader.arguments import Arguments
 
 
+
 def parse_arguments(args_in) -> Arguments:
     """
     Configures the command-line interface.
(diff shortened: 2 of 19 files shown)


src/lms-ds-loader/README.md
```diff
@@ -34,9 +34,16 @@ poetry run python ./lms_ds_loader/main.py --server localhost --dbname fizz --use
 
 ## Dev Operations
```
bradbanister, line 35: I like this style for conveying script information.

src/lms-ds-loader/README.md
```diff
@@ -34,9 +34,16 @@ poetry run python ./lms_ds_loader/main.py --server localhost --dbname fizz --use
...
-1. View coverage report in the console: `poetry run coverage report`
-1. Generate an XML coverage report: `poetry run coverage xml` generates file
-   `coverage.xml`
+1. View code coverage:
+    * View coverage report in the console: `poetry run coverage report`
+    * Generate an XML coverage report: `poetry run coverage xml` generates file
+      `coverage.xml`
+    * Generate an HTML file to view in a browser: `poetry run coverage html`
+      generates file  `htmlcov/index.html`. *This version is clickable*,
+      allowing drilldown to see coverage details.
```
lalrg, line 49: We could add this section to all the readme files, since now we are having similar project structures

src/lms-ds-loader/lms_ds_loader/argparser.py
```diff
@@ -23,7 +23,7 @@ def parse_arguments(args_in) -> Arguments:
         A populated `Arguments` object.
     """
 
-    assert args_in is not None
+    assert type(args_in) is list, "Argument `args_in` must be a list"
```
bradbanister, line 26: Didn't know about the message possibility, will have to add to my stuff.

src/lms-ds-loader/lms_ds_loader/arguments.py
```diff
@@ -71,9 +47,18 @@ def set_connection_string_using_integrated_security(self, server, port, db_name)
         db_name : str
             Database name.
         """
+        assert type(server) is str, "Argument `server` must be a string"
```
bradbanister, line 50: FYI, `assert isinstance(server, str)` also works.  Seems to be community debate on which is better for primitive types.

src/lms-ds-loader/lms_ds_loader/arguments.py
```diff
@@ -18,45 +22,17 @@ class Arguments:
...
 
     @staticmethod
-    def _build_for_mssql(server, port, db_name, username, password):
-        assert server is not None
-        assert server.strip() != ""
-
-        assert db_name is not None
-        assert db_name.strip() != ""
-
+    def _get_mssql_port(port):
```
lalrg, line 29: Reading that this static method starts with _ I am assuming it should be internal or not intended to be used outside of the function. I'd put this function outside of the class... So when someone imports the Arguments class, won't be able to do Arguments._get_mssql_port() from outside the class.

src/lms-ds-loader/lms_ds_loader/arguments.py
```diff
@@ -93,5 +78,31 @@ def set_connection_string(self, server, port, db_name, username, password):
             Database password.
         """
 
+        assert type(server) is str, "Argument `server` must be a string`"
```
lalrg: It would be nice defining a single place for error messages, then importing it and replacing values where needed. So we can avoid hardcoding repetitive strings in the code.

src/lms-ds-loader/lms_ds_loader/mssql_lms_operations.py
```diff
@@ -0,0 +1,196 @@
+# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: Could this be simplified by dropping the dataclass and just having functions in the module?

bradbanister (approved):
Just some comments and one suggestion

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: If that works, might be better to use the `isinstance` just for consistency. Easy enough for me to try since I have these all unit tested 😀 .

stephenfuqua: All tests pass, as expected, with change to `isinstance`.

src/lms-ds-loader/lms_ds_loader/mssql_lms_operations.py
stephenfuqua: There are other classes where that might work, but for this one no: the class is instantiated early on, after reading the CLI input arguments, and passed from function to function. In this way we can pass around either an MSSQL object or a PostgreSQL object.

lalrg (approved):

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: Not intended for use outside of this class. But it is not critical to _prevent_ it from being used.  I can see that it is possible to move it out of the class. I tried that and all tests pass. 

Where does it belong though? If it is truly related to "arguments", even if static, then I think it belongs with the class. But otherwise, it could just be a bare function in the module. In this case, it is hard to tell - just not very clear cut. 

I'm inclined to leave it alone unless anyone else has a strong preference to move it out of the class.

src/lms-ds-loader/lms_ds_loader/arguments.py
stephenfuqua: That's true, I am rewriting that same text over and over. That may be worthwhile.

src/lms-ds-loader/lms_ds_loader/arguments.py
bradbanister: I probably would have put at the module level, but no strong preference.

PR #41, opened 2020-09-30
[FIZZ-85] Added MetaEd documentation and updated SQL scripts

ggaronzik:


bradbanister:
@ggaronzik  Is the full MetaEdOutput folder supposed to be there?

ggaronzik:
> @ggaronzik Is the full MetaEdOutput folder supposed to be there?

We initially suppressed the full folder but then I believe @stephenfuqua made a modification to the .gitignore. We may need to update this based on what files we want to see. Or since I've moved the SQL scripts to a new directory, we could remove the folder altogether again.

docs/metaed-project/Association/UserLMSSectionAssociation.metaed
```diff
@@ -1,37 +1,37 @@
-﻿// SPDX-License-Identifier: Apache-2.0
-// Licensed to the Ed-Fi Alliance under one or more agreements.
-// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
-// See the LICENSE and NOTICES files in the project root for more information.
-
```
stephenfuqua, line 5: @ggaronzik is it possible to move this license boilerplate back to the top of the file instead of being a footer?

docs/metaed-project/DomainEntity/User.metaed
```diff
@@ -1,38 +1,38 @@
-﻿// SPDX-License-Identifier: Apache-2.0
-// Licensed to the Ed-Fi Alliance under one or more agreements.
-// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
-// See the LICENSE and NOTICES files in the project root for more information.
-
 Domain Entity User
```
stephenfuqua, line 1: I hesitate to ask this... but after writing SQL scripts with `User`, I've realized the annoyance of having to write `[User]` everywhere since "user" is a reserved word. Using the square bracket really isn't a big deal - except that it is not portable to PostgreSQL, so any queries with the square brackets must be rewritten to remove them for PostgreSQL. So... how difficult would it be to change this to `LMSUser` instead of `User`?

docs/metaed-project/MetaEdOutput/EdFi/Database/PostgreSQL/ODS/Structure/0040-IdColumnUniqueIndexes.sql
```diff
@@ -0,0 +1,192 @@
+CREATE UNIQUE INDEX IF NOT EXISTS UX_a97956_Id ON edfi.AcademicWeek(Id);
```
stephenfuqua, line 1: Is there anyway to build the MetaEd output without creating the `edfi` schema and all of its tables?

stephenfuqua (changes requested):

docs/metaed-project/MetaEdOutput/EdFi/Database/PostgreSQL/ODS/Structure/0040-IdColumnUniqueIndexes.sql
bradbanister: It's all-or-nothing per database.

stephenfuqua: I will look at the filesystem to see if we can gitignore the "edfi" artifacts while keeping the "fizz" artifacts.

docs/metaed-project/Association/UserLMSSectionAssociation.metaed
ggaronzik: MetaEd does not expect any text, even comments, before the initial declaration. So I moved these all to the bottom. I could have them be on the second line but I could build with the comments at the top. @bradbanister may have an alternative suggestion here.

docs/metaed-project/DomainEntity/User.metaed
ggaronzik: I don't think this is a big lift. I will need to update the User MetaEd file and any references, the UDM, and the documentation in techdocs. I would also need to recreate the manual changes to the supporting-artifacts for the Identity fields.

@ejansson, do you see any issues with making this change?

docs/metaed-project/Association/UserLMSSectionAssociation.metaed
bradbanister: @ggaronzik  I'm a bit surprised MetaEd doesn't like it.  Can you open a bug?

docs/metaed-project/DomainEntity/User.metaed
stephenfuqua: We could do that in a separate ticket.

docs/metaed-project/Association/UserLMSSectionAssociation.metaed
stephenfuqua: Somehow this isn't a problem in the main ODS codebase, for instance: https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/main/Extensions/Homograph/HomographMetaEd/Association/StudentSchoolAssociation.metaed

stephenfuqua: But we can certainly live with having the text at the bottom of the file for now.

stephenfuqua:
@ggaronzik it looks like you removed the wrong `.gitignore` file, as the HTML documentation and the Excel spreadsheets are back now. Can I fix up this branch so that my `.gitignore` file is restored and we're only keeping the SQL output for the "fizz" extension?

docs/metaed-project/Association/UserLMSSectionAssociation.metaed
ggaronzik: Comments no longer a bug - these have been moved to the top of the files again with the latest commit.

docs/metaed-project/DomainEntity/User.metaed
ggaronzik: Updated User to LMSUser.

ggaronzik:
> @ggaronzik it looks like you removed the wrong `.gitignore` file, as the HTML documentation and the Excel spreadsheets are back now. Can I fix up this branch so that my `.gitignore` file is restored and we're only keeping the SQL output for the "fizz" extension?

@stephenfuqua, I believe the last of the changes for this ticket are in so you can make the modifications for the .gitignore file.

stephenfuqua (approved):

PR #43, opened 2020-10-01
[FIZZ-49] Add pagination support to Schoology scripts

lalrg:


src/schoology-extractor/schoology_extractor/api/paginated_result.py
```diff
@@ -0,0 +1,113 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+from __future__ import annotations
```
lalrg, line 6: This import is required in order to be able to annotate a return type of the class itself:
`def get_next_page(self) -> PaginatedResult`

src/schoology-extractor/schoology_extractor/api/request_client_base.py
```diff
@@ -0,0 +1,87 @@
+# SPDX-License-Identifier: Apache-2.0
```
lalrg, line 1: Created this base class for not having circular dependencies between RequestClient and PaginatedResult:
Without the base class:
RequestClient returns PaginatedResult # In order to annotate this, I needed to import PaginatedResult  inside of RequestClient
PaginatedResult needs a RequestClient # For get_next_page method

With the base class:
RequestClient(RequestClientBase) # Imports only RequestClientBase
PaginatedResult needs a RequestClientBase # Only needs access to base get method(not get_users or any other method)

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -21,27 +21,39 @@
 
 
 # export users
-export_data.to_csv(request_client.get_users(), f"{schoology_output_path}/users.csv")
+users_response = request_client.get_users()
+users_list = []
+while True:
+    users_list = users_list + users_response.current_page_items
+    if users_response.get_next_page() is None:
```
lalrg, line 28: users_response.get_next_page() returns self when there is another page, and returns None when is the last page.

src/schoology-extractor/schoology_extractor/api/request_client_base.py
bradbanister: interesting, haven't run into this dependency issue yet, but i can see it happening.

bradbanister:
Since you are using type annotations, you should run Pylance on your code base.  Several annotations are wrong.

src/schoology-extractor/schoology_extractor/api/paginated_result.py
```diff
@@ -0,0 +1,113 @@
...
+    def get_next_page(self) -> PaginatedResult:
+        """
+        Send an HTTP GET request for the next page.
+
+        Returns:
+            PaginatedResult: If there are more pages.
+            None: If there are not more pages.
+        """
+        if "links" not in self._api_response:
+            return None
```
bradbanister, line 95: The type annotation for this method is wrong if you are allowing None to be returned.  It should be Optional[PaginatedResult]

src/schoology-extractor/tests/api/test_paginated_result.py
```diff
@@ -0,0 +1,121 @@
...
+
+            # Assert
+            with pytest.raises(AssertionError):
+                PaginatedResult(None, None, None, None, None)
+
+        def test_given_wrong_types_in_parameters_shoud_throw_assertException(self):
+
+            # Assert
+            with pytest.raises(AssertionError):
+                PaginatedResult(0, 1, 'None', [], '')
```
bradbanister, line 53: @stephenfuqua @lalrg  I'm not seeing the cost/benefit of these wrong type tests if we are going to use type annotations, which I think we should.  For example, here and on line 47 the Pylance extension highlights them as errors in VS Code.

If you want to keep them around you'll need use the standard `# type: ignore` to suppress the errors.

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -23,59 +22,21 @@ class RequestClient:
...
+        base_url: str = DEFAULT_URL
+    ):
+        assert schoology_key is not None
+        assert schoology_secret is not None
+        assert base_url is not None
 
+        super().__init__(schoology_key, schoology_secret, base_url)
 
-    def get_assignments_by_section_ids(self, section_ids):
+    def get_assignments_by_section_ids(self, section_ids: list) -> dict:
```
bradbanister, line 39: You're returning a list, not a dict

src/schoology-extractor/schoology_extractor/api/request_client_base.py
bradbanister: @stephenfuqua @lalrg  The more I look at this the less I like it.  Fortunately I found an answer:
[Python type hints with imported class methods
](https://stackoverflow.com/questions/63234592/python-type-hints-with-imported-class-methods)
Basically you avoid this by guarding the type-only import with a "TYPE_CHECKING" constant:
```
from __future__ import annotations
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from .request_client import RequestClient
```

bradbanister (changes requested):

src/schoology-extractor/schoology_extractor/api/request_client_base.py
bradbanister: TypeScript solves this with the "import type" variation of import.

src/schoology-extractor/tests/api/test_paginated_result.py
stephenfuqua: I have yet to see any advantage to type annotations other than to the reader. Even if Pylance flags something in the IDE - which it isn't doing for me (I'll do some reading to try to understand why) - will that prevent something strange from happening at runtime?

Avoiding the equivalent of null reference errors, and avoiding accidental empty strings, is important to me for preventing surprise bugs, the kinds of bugs that I have seen and had difficulty debugging in the past. So I would rather throw an error proactively then let them bubble up from deeper in the code, so that I can have better control over the messaging and understanding what is actually occurring.

I also have no idea what this comment about `# type: ignore` is referring to.

bradbanister: Splitting my comments up here:
1) My IDE experience with type annotations has been fantastic.  It's equivalent to the difference between writing plain Javascript and writing TypeScript.  I strongly advocate for their use and am happy to help with any set up issues.
2) No argument with the usage of assert in the code base.
3) However, I see little value in writing tests that only highlight the same issues that a type checker will highlight.
4) If we want to write those tests anyway, put a `# type: ignore` comment on the end of the offending line.  In this example, line 53.

src/schoology-extractor/tests/api/test_request_client_base.py
```diff
@@ -0,0 +1,93 @@
...
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import pytest
+
+from schoology_extractor.api.request_client_base import RequestClientBase
+
+fake_key = 'TEST_KEY'
+fake_secret = 'TEST_SECRET'
+fake_endpoint_url = 'FAKE_URL'
+default_url = 'https://api.schoology.com/v1/'
```
stephenfuqua, line 13: https://www.python.org/dev/peps/pep-0008/#constants

> "Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL."

stephenfuqua (changes requested):

stephenfuqua:
@lalrg I also noticed that you've changed from grading periods to section_ids in the environment variables. Can you tell me more about that?

I must have missed some conversation about why this was changed. I would think that listing out a few grading periods would be much simpler than listing out all of the sections. For example, a school would likely be interested in less than 5 grading periods at a time (I'm guessing), but might have hundreds of sections.

The readme still refers to grading periods and a script that does not exist in this directory, `grading_periods.py`.

lalrg:
> @lalrg I also noticed that you've changed from grading periods to section_ids in the environment variables. Can you tell me more about that?
> 
> I must have missed some conversation about why this was changed. I would think that listing out a few grading periods would be much simpler than listing out all of the sections. For example, a school would likely be interested in less than 5 grading periods at a time (I'm guessing), but might have hundreds of sections.
> 
> The readme still refers to grading periods and a script that does not exist in this directory, `grading_periods.py`.

@stephenfuqua There is no reason for this change or at least I don't remember, probably at some point in development I somehow changed the parameter. I can refactor if needed

stephenfuqua:
> @stephenfuqua There is no reason for this change or at least I don't remember, probably at some point in development I somehow changed the parameter. I can refactor if needed

Yes, please restore.

src/schoology-extractor/schoology_extractor/api/paginated_result.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(
+        self,
+        request_client: "RequestClient",
+        page_size: int,
+        api_response: dict,
+        resource_name: str,
+        requested_url: str,
+        current_page: int = 1,
+    ):
+        assert hasattr(request_client, "get"), "Argument `request_client` should be of type `RequestClient`."
```
lalrg, line 46: I had to change this assert in order to remove the circular import error

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -13,35 +13,66 @@
...
+        break
 
+export_data.to_csv(users_list, f"{schoology_output_path}/users.csv")
```
bradbanister, line 37: Suggest replacing these with [os.path.join()](https://docs.python.org/3/library/os.path.html#os.path.join)

bradbanister (changes requested):
I get an empty assignments.csv when I run with
`SCHOOLOGY_GRADING_PERIODS=2942191527,2975852079,2941242697`
Is that expected?

Also, I don't see the page size setting mentioned in Steven's test summary [here](https://tracker.ed-fi.org/browse/FIZZ-49?focusedCommentId=58804&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-58804)

bradbanister (approved):
Discussed how there is no need for page size in .env

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -190,8 +194,12 @@ def get_users(self, page_size: int = 20) -> PaginatedResult:
...
-        url = f"users?start=0&limit={page_size}"
+        url = f"users?{self._build_pagination_params(page_size)}"
 
         return PaginatedResult(
             self, page_size, self.get(url), "user", self.base_url + url
         )
+
+    def _build_pagination_params(self, page_size: int):
+        assert isinstance(page_size, int)
+        return f"start=0&limit={page_size}"
```
stephenfuqua, line 268: How do you get the second page, if the `start` is always zero?

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -50,7 +56,9 @@
 
 for section_id in sections_id_array:
     submissions_response = request_client.get_submissions_by_section_id(section_id)
+
     while True:
+        print(submissions_response.current_page_items)
```
stephenfuqua: For the logging story: good example of something that might be a DEBUG log entry.

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -144,19 +148,31 @@ def get_section_by_id(self, section_id: str) -> dict:
...
 
         """
-        assert isinstance(section_id, str), "Argument `section_id` should be of type `str`."
+        assert isinstance(course_ids, list), "Argument `course_ids` should be of type `list`."
+
+        sections = []
+        for course_id in course_ids:
+            url = f"courses/{course_id}/sections"
+            sections_per_course = PaginatedResult(
+                self, 20, self.get(url), "section", self.base_url + url
```
stephenfuqua, line 157: Is this hard-coding a page size of 20?

src/schoology-extractor/schoology_extractor/helpers/export_data.py
```diff
@@ -6,16 +6,43 @@
...
     """
-    assert data is not None
-    assert output_path is not None
+    assert isinstance(data, list)
+    assert isinstance(output_path, str)
 
     df = pd.DataFrame(data)
 
     df.to_csv(output_path, index=False)
     print(f"The file has been generated => {output_path}")
```
stephenfuqua, line 27: I would probably log this at the INFO level.

src/schoology-extractor/schoology_extractor/main.py
stephenfuqua: It is probably a good idea. Personally I've been very lax about that, since PowerShell is perfectly happy to accept forward slashes. However, you're right that it _is_ a better practice: with the forward slash, we're _forcing_ people to use PowerShell instead of `cmd.exe`.

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -36,29 +36,40 @@
...
+# now we can get a list of sections
+course_ids = map(lambda x: x["id"], courses_list)
+sections_list = request_client.get_section_by_course_ids(list(course_ids))
 export_data.to_csv(sections_list,
                    f"{schoology_output_path}/sections.csv")
 
 
 # export assigments
-export_data.to_csv(request_client.get_assignments_by_section_ids(sections_id_array),
+section_ids = map(lambda x: x["id"], sections_list)
```
stephenfuqua, line 57: This code seems fine, but I'm curious why you chose `map` instead of writing a list comprehension?

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -208,6 +208,7 @@ def get_users(self, page_size: int = 20) -> PaginatedResult:
             A parsed response from the server
 
         """
+        assert isinstance(page_size, int), "Argument `page_size` should be of type `str`."
```
stephenfuqua, line 211:
```suggestion
        assert isinstance(page_size, int), "Argument `page_size` should be of type `int`."
```

src/schoology-extractor/schoology_extractor/api/request_client.py
lalrg: Yes, in some situations where we have complex requests... Such as needing to merge the results from different paginated results, I opted for just returning the final list and not the paginated result.

src/schoology-extractor/tests/api/test_request_client.py
```diff
@@ -228,4 +233,34 @@ def test_given_correct_parameter_then_url_is_built_correctly(self, default_reque
...
+        def test_given_correct_parameter_then_get_call_is_made(self, default_request_client, mocker):
+            # Arrange
+            get_mock = mocker.MagicMock(return_value={})
+            default_request_client.get = get_mock
+
+            # Act
+            default_request_client.get_grading_periods()
+
+            # Assert
+            assert get_mock.call_count == 1
```
stephenfuqua, line 250: Wouldn't it be more valuable to ensure that the expected results are returned, rather than only confirming that the method was called?

src/schoology-extractor/schoology_extractor/main.py
lalrg: No specific reason... Normally I use list comprehension only when I need to filter something, but any of them would be fine.

src/schoology-extractor/tests/helpers/test_export_data.py
```diff
@@ -29,5 +30,18 @@ def test_then_call_DataFrame_method(self, mocker):
...
+        def test_then_call_DataFrame_method(self, mocker):
 
+            # Arrange
+            DataFrame_mock = mocker.patch.object(pd, "DataFrame")
+
+            # Act
+            to_string([])
+
+            # Assert
+            DataFrame_mock.assert_called_once()
```
stephenfuqua, line 47: This is an example of over-specification in a test, which is one of the dangers of mocking in general. It would be preferable to define the expected output and not mock the DataFrame. Just confirm that the actual output matches the expected output. Then if someone wants to change the code to get the expected output without pandas, they don't need to change the unit test as well.

src/schoology-extractor/schoology_extractor/main.py
stephenfuqua: It is not worth changing this code. In general I recommend sticking with list comprehensions even when not filtering something, because it is strange to see a mixture of the two and slows down my reading of the code.

Interesting performance comparison: https://stackoverflow.com/a/1247490/30384

src/schoology-extractor/schoology_extractor/api/request_client.py
stephenfuqua: I admit I don't understand that. Just looking at the code, it looks like a bug. This might be a good time to replace `20` with a well-named constant (https://refactoring.guru/replace-magic-number-with-symbolic-constant), and possibly add a text comment explaining why this is being done.

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -8,35 +8,53 @@
...
 
-    Args:
-        schoology_key (str): The consumer key given by Schoology
-        schoology_secret (str): The consumer secret given by Schoology
-        base_url (str, optional): The API base url. Default value: https://api.schoology.com/v1/
+    Attributes
+    ----------
+    oauth : OAuth1Session
+        The two-legged authenticated OAuth1 session.
+    '''
```
stephenfuqua, line 35: I had not previously noticed the ability to describe attributes like this. Interesting.

src/schoology-extractor/schoology_extractor/api/request_client.py
lalrg: This parameter is only called the first time, for the next requests it will get the next page link from the Schoology's response

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -74,61 +102,167 @@ def _get(self, url):
         )
         return response.json()
```
stephenfuqua, line 103: What happens if `requests-oauthlib` encounters an HTTP response code of 4xx or 5xx? I recommend that you do some experimentation to see if you can trigger 4xx response. Triggering a 5xx response would be more difficult. See how this library and the main request library handle non-2xx error codes. Whatever they do for 4xx may be similar to 5xx - but we might need to build a quick-and-dirty web service that returns 500 responses so that we can test again that and make sure we understand the response. I want to make sure have good messages for the user. We also need to think about retries at that point.

I'm going to create a separate ticket for this point. I don't want to delay this ticket with this additional scope, and I don't think this scope should go into the logging ticket. No need to make any changes right now.

src/schoology-extractor/schoology_extractor/api/request_client.py
lalrg: I will

src/schoology-extractor/schoology_extractor/api/request_client.py
stephenfuqua: Ok, that makes good sense now. Suggestion: update the function name so the reader will understand this clearly. Perhaps `_build_query_params_for_first_page`?

FYI, generally I do not write argument validations for "private" methods. Why? It is a compromise: instead of saying that _all_ functions must validate _all_ arguments, I usually ask that _public_ (and _protected_ in C#/Java/etc) functions need argument validation. With private functions, and a small class, it is usually easy to see that the values are already validated elsewhere.

So, there's no reason to delete the validation right now. Just letting you know that I would not have complained if it had not been there.

src/schoology-extractor/schoology_extractor/api/request_client.py
stephenfuqua: There's already a ticket for that 😆 . FIZZ-50.

lalrg: It fails in this line when response is empty(not 200)... I think we have a ticket for error handling: https://tracker.ed-fi.org/browse/FIZZ-50

src/schoology-extractor/schoology_extractor/api/request_client.py
lalrg: Renamed ✅

src/schoology-extractor/schoology_extractor/api/request_client.py
lalrg: ✅

src/schoology-extractor/schoology_extractor/helpers/export_data.py
lalrg: Sounds good... That's the next story

src/schoology-extractor/schoology_extractor/api/request_client.py
lalrg: ✅

src/schoology-extractor/tests/helpers/test_export_data.py
lalrg: ✅

src/schoology-extractor/tests/api/test_request_client.py
lalrg: You are right, I'll fix it

lalrg: ✅

lalrg: Done ✅

src/schoology-extractor/tests/helpers/test_export_data.py
lalrg: Changed ✅

stephenfuqua (approved):
I approve for my part, though leaving the final approval and merge to @bradbanister.

bradbanister (approved):
I approve as well.  Actually thought I approved yesterday.

PR #45, opened 2020-10-02
[FIZZ-52] Build automation for Schoology Extractor

stephenfuqua:
I've created a "universal" build script that can be called for any utility. Of course some commands may fail if the utility does not have the proper package installed - for example, the "typecheck" command will fail if `mypy` is not installed into the utility.

Why?

Because it keeps me from having to remember each command's nuances, especially when integrating with a CI tool. Note that it relies only on python core libraries, and thus does not itself need to run inside of a virtual environment.


.teamcity/_Self/Schoology/buildTypes/TestBranch.kt
```diff
@@ -0,0 +1,17 @@
...
+import jetbrains.buildServer.configs.kotlin.v2019_2.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.swabra
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildSteps.powerShell
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.VcsTrigger
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.vcs
+
+object TestBranch : BuildType ({
+    name = "Branch Testing"
+    templates(_self.templates.TestTemplateBranch)
+})
```
stephenfuqua, line 17: Note that I did not need to do any customization in this BuildType: it only needs to apply the required template.

.teamcity/_Self/templates/TestTemplateBase.kt
```diff
@@ -0,0 +1,81 @@
...
+
+package _self.templates
+
+import jetbrains.buildServer.configs.kotlin.v2019_2.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.freeDiskSpace
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.swabra
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildSteps.script
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.vcs
+
+open class TestTemplateBase : Template({
```
stephenfuqua, line 14: This is the abstract base class for other templates.

.teamcity/_Self/templates/TestTemplateBase.kt
```diff
@@ -0,0 +1,81 @@
...
+    name = "Python Test Template Base Class"
+    id = RelativeId("TestTemplateBase")
+
+    artifactRules = "+:%project.directory%/htmlcov => coverage.zip"
```
stephenfuqua, line 18: The Python coverage project outputs coverage results in formats that don't work directly in TeamCity. There's probably a plugin somewhere that supports the Java Cobertura code coverage format used by the python coverage XML reporter... but I didn't find it and it isn't important enough. I want to be able to see the code coverage, but I'm not going to fail a build because of code coverage. We don't do that in our other Ed-Fi projects. Therefore an HTML report is appropriate. Here in this artifact rule we simply create a zip file of the `htmlcov` directory that is created by the coverage utility.

.teamcity/_Self/templates/TestTemplateBase.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    steps {
+        script {
+            name = "Install"
+            formatStderrAsError = true
+            workingDir = "eng"
+            executionMode = BuildStep.ExecutionMode.RUN_ON_SUCCESS
+            scriptContent = """
+                    python ./build.py install ../%project.directory%
+                """.trimIndent()
```
stephenfuqua, line 33: Here's why I created a Python script to help automate the build process: this is not Windows specific. This TeamCity build configuration should run on Linux as well. Not that we're trying to run it there right now, but it is nice to know that we have that flexibility.

.teamcity/_Self/templates/TestTemplateBase.kt
```diff
@@ -0,0 +1,81 @@
...
+            executionMode = BuildStep.ExecutionMode.RUN_ON_SUCCESS
+            scriptContent = """
+                    python ./build.py coverage:html ../%project.directory%
+                """.trimIndent()
+        }
+        script {
+            name = "Type Check"
+            formatStderrAsError = true
+            workingDir = "eng"
+            executionMode = BuildStep.ExecutionMode.RUN_ON_FAILURE
```
stephenfuqua, line 48: I decided to let this "run on failure" (and the next step) so that, even if there are unit test failures, we can still get the type and style checks. I want the unit test failures to mark the whole build as a failure, but it is nice to still have these two other steps run anyway.

.teamcity/_Self/templates/TestTemplateBranch.kt
```diff
@@ -0,0 +1,26 @@
...
+// See the LICENSE and NOTICES files in the project root for more information.
+
+package _self.templates
+
+import jetbrains.buildServer.configs.kotlin.v2019_2.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.VcsTrigger
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.vcs
+
+object TestTemplateBranch : TestTemplateBase() {
+    init {
```
stephenfuqua, line 13: Note that this inherits from the base template class, and adds a trigger on the `main` branch.

.teamcity/_Self/templates/TestTemplatePullRequest.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import jetbrains.buildServer.configs.kotlin.v2019_2.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.commitStatusPublisher
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.PullRequests
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.pullRequests
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.VcsTrigger
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.vcs
+
+object TestTemplatePullRequest : TestTemplateBase() {
+    init {
```
stephenfuqua, line 16: This inherits from the template base class, adds a trigger on pull requests, and then adds two important features.

1. "Commit status publisher" is what pushes the TeamCity status information out to GitHub.
2. "Pull requests" is a tool that improves the pull request build process: it only automates pull requests into the `main` branch (default), and it limits the builds to only people who are members or collaborators on the project. That is, some random person can't create a pull request and have it trigger a build, potentially with malicious code in it.

eng/build.py
```diff
@@ -0,0 +1,178 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
```
stephenfuqua, line 4: In the next story about build automation, I will add an acceptance testing criteria suggesting that this script be documented in the main readme, or in a developer readme, so that developers know how to use it for localhost command line operations.

eng/build.py
```diff
@@ -0,0 +1,178 @@
...
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+"""Build automation scripts"""
+
+import os
+import subprocess
+import sys
+from typing import List
```
stephenfuqua, line 11: Note that this script does not depend on any external libraries and does not need to run in a virtual environment.

eng/build.py
```diff
@@ -0,0 +1,178 @@
...
+    if exit_immediately:
+        exit(result.returncode)
+
+    if result.returncode != 0:
+        exit(result.returncode)
+
+
+def _run_install():
+    _run_command([
+        "poetry",
```
stephenfuqua, line 57: OK, actually it does depend on `poetry` being installed already!

eng/build.py
lalrg: It would be nice if we at least mention the python version that we're using when running this script, or even better to specify the minimum python version.

lalrg: Ok I saw the validation for Python 3.8, but we could add it to the .teamcity readme

eng/build.py
```diff
@@ -0,0 +1,178 @@
...
+
+"""Build automation scripts"""
+
+import os
+import subprocess
+import sys
+from typing import List
+
+
+def _display_help():
```
lalrg, line 14: I'd just use the docstrings for displaying help... You can display docstrings in console by running `print(__doc__)`

.teamcity/readme.md
```diff
@@ -0,0 +1,50 @@
+# Getting Started
+
+To apply these build configuration settings on a new TeamCity environment (min
+version: `2019.2.x`):
```
stephenfuqua, line 4:
```suggestion
version: TeamCity - `2019.2.x`, Python - `3.8.x`, Poetry - 1.0.x):
```

eng/build.py
stephenfuqua: Done.

eng/build.py
stephenfuqua: I had not thought about putting this information into a docstring. Let's consider that when making the next round of improvements, while working on build automation for Google Classroom.

lalrg: Oh, since it is not a class, it could be harder to display the entire docstring from a function... I think it is okay, from that perspective.

lalrg (approved):
Looks good to me

PR #52, opened 2020-10-08
[FIZZ-96] Google Classroom assignment UDM generation

bradbanister:


bradbanister:
This adds UDM assignment generation.  I plan to increase overall project test coverage as part on the next story, generating UDM submissions.

src/google-classroom-extractor/google_classroom_extractor/mapping/assignments.py
```diff
@@ -0,0 +1,132 @@
...
+    assert "dueTime.hours" in coursework_df.columns
+    assert "dueTime.minutes" in coursework_df.columns
+
+    if {
+        "dueDate.year",
+        "dueDate.month",
+        "dueDate.day",
+        "dueTime.hours",
+        "dueTime.minutes",
+    }.issubset(coursework_df.columns):
```
stephenfuqua, line 67: Don't the asserts statements above guarantee that this `if` condition is `True`?

src/google-classroom-extractor/google_classroom_extractor/mapping/assignments.py
```diff
@@ -0,0 +1,132 @@
...
+            raise ValueError  # fillna will never return None in this usage
+        coursework_df["DueDateTime"] = filled_df.apply(
+            lambda date_element: datetime(*pd.to_numeric(date_element, downcast="integer")),
+            axis=1,
+        )
+    else:
+        coursework_df["DueDateTime"] = ""
+
+    coursework_df["SourceSystemIdentifier"] = coursework_df[["courseId", "id"]].agg(
+        ":".join, axis=1
```
stephenfuqua, line 88: That's a nice technique, I wouldn't have thought to use `agg`. I probably would have used `apply`. This is more concise and future proof (additional columns) than

```
.apply(lambda row: f"{row['courseId']}:{row['id']}")
```

src/google-classroom-extractor/tests/mapping/assignments/test_assignments.py
```diff
@@ -0,0 +1,214 @@
...
+    assignments_dfs: Dict[str, pd.DataFrame] = coursework_to_assignments_dfs(
+        coursework_df
+    )
+
+    # assert
+    assert len(assignments_dfs) == 1
+    assignments_df: pd.DataFrame = assignments_dfs[COURSE_ID]
+    row_count, column_count = assignments_df.shape
+    assert row_count == 1
+    assert column_count == 12
```
stephenfuqua, line 81: I'm OK with leaving this for now. It might be nice for us to look into using equivalents to NUnit's [SetUp] and related functions to structure something where each test can have a single assert. That way if the assert on line 80 fails, we still run the test that has the assert on line 81.

stephenfuqua (approved):

src/google-classroom-extractor/google_classroom_extractor/mapping/assignments.py
bradbanister: Yikes, good catch. They are optional.

src/google-classroom-extractor/tests/mapping/assignments/test_assignments.py
bradbanister: Yes, I was thinking the same thing.  Wasted a good chunk of time trying to figure out the PyTest way. Let me know if you find anything.

src/google-classroom-extractor/google_classroom_extractor/mapping/assignments.py
stephenfuqua: In slack we agreed to handle this in the next story.

PR #56, opened 2020-10-13
[FIZZ-50] Retry logic and exception handling

stephenfuqua:
I created a [quick-and-dirty API test harness](https://github.com/Ed-Fi-Alliance/ed-fi-gist/tree/stephen.fuqua/fake-api) in JavaScript - took me only a few minutes after struggling to get a Python-based simple service running. Can use that in combination with the `SCHOOLOGY_BASE_URL` environment variable to inject different HTTP status codes into the response. Although the `requests` module handles these status codes in standard ways, it was useful to see the results in action and for testing that our code now handles these different statuses.

The "service layer" class - best name I could up with with for this business logic - is now 97% tested.

The application is running correctly and this is now ready for QA.


src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -27,115 +27,72 @@
...
+    assignments = facade.get_assignments(closure["sections"], grading_periods)
+    closure["assignments"] = assignments
+    return assignments
+
+
+def _export_submissions():
+    return facade.get_submissions(closure["assignments"])
+
+
+def main():
```
lalrg, line 89: Instead of using a closure object, I'd create just pure functions that return values, storing the results in simple variables and passing them as parameters where needed.
My only reason for this change, would be to keep things as declarative as possible.
Example:
Method `_export_sections()` is not exporting sections, it is returning a list of sections from the API and setting a value in the closure object. In case you wanted to keep that closure, I'd rename the function to something similar to `_get_sections_and_set_values_in_closure()`, keeping in mind that the function is actually doing two things: setting a value in the object, and returning a value that it gets from the API.

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -3,21 +3,32 @@
...
+from opnieuw import retry
+from requests.exceptions import ConnectionError, HTTPError, Timeout
+from requests.packages.urllib3.exceptions import ProtocolError  # type: ignore
 from requests_oauthlib import OAuth1Session  # type: ignore
 
 from .paginated_result import PaginatedResult
 
 
-DEFAULT_URL = "https://api.schoology.com/v1/"
+DEFAULT_URL = os.environ.get("SCHOOLOGY_BASE_URL") or "https://api.schoology.com/v1/"
```
lalrg, line 20: Is this a value that we can pass as an arg/.env? or it is only going to be configurable from .env?

lalrg: Same question for REQUEST_RETRY_COUNT and REQUEST_RETRY_TIMEOUT_SECONDS

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -123,15 +151,21 @@ def get_assignments_by_section_ids(self, section_ids: list) -> list:
...
 
         """
-        assert isinstance(section_ids, list), "Argument `section_ids` should be of type `list`."
+        assert isinstance(
+            section_ids, list
+        ), "Argument `section_ids` should be of type `list`."
 
         assignments: List[object] = []
         for section_id in section_ids:
             params = self._build_query_params_for_first_page(DEFAULT_PAGE_SIZE)
```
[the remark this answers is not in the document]
stephenfuqua: done

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -123,15 +151,21 @@ def get_assignments_by_section_ids(self, section_ids: list) -> list:
...
             url = f"sections/{section_id}/assignments?{params}"
 
             assignments_per_section = PaginatedResult(
-                self, DEFAULT_PAGE_SIZE, self.get(url), "assignment", self.base_url + url
+                self,
+                DEFAULT_PAGE_SIZE,
```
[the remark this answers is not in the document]
stephenfuqua: I added the parameter but then forgot to use it 😆 .

src/schoology-extractor/schoology_extractor/main.py
stephenfuqua: I agree about renaming the methods - they're not doing the exporting.  I suppose I could simplify - instead of passing a Callable around, I could make the call and pass the data to the `_create_file...` functions. But I still have the problem of having the try/catch only once. I will think about it.

src/schoology-extractor/schoology_extractor/api/request_client.py
stephenfuqua: I see these as changing rarely, so I didn't want to make them into full input parameters - especially the URL. But I did want a way to introduce different values in rare circumstances.

src/schoology-extractor/schoology_extractor/main.py
stephenfuqua: Basically I'm following the strategy pattern by injecting the callable into the shared code that performs the actual export and wraps both the get and save with try/catch. But I needed to be able to pass some data around. The "closure" variable is scoped to the entire script and is thus immutable inside the function. While you can't set the value of that outer variable, you _can_ add entries to it. So `closure` is kind of a global container.

stephenfuqua: @bradbanister would love to have your thoughts on this as well.

bradbanister:
For doing a fake API server, take a look at [Pretenders](https://pretenders.readthedocs.io/en/latest/).  It's Python, and the example shows how you can spin up and configure a server right in your PyTests.  It looks like the project hasn't been updated in a while, but I don't see that as an issue for this kind of testing tool.

src/schoology-extractor/schoology_extractor/main.py
bradbanister: For the closure object, it's similar to my "Result" object.  I'd give the closure object an actual type (I used a simple namedtuple) so mypy can validate fields.  I'd also rename to something more expressive...like "Result" 😉, or maybe ApiResults, etc.

For the facade, it looks like the primary purpose is around paging.  I'd again rename to something more expressive, maybe PagingDecorator?  I also wonder if it would be nicer as an actual Python decorator, meaning convert most of it to a function wrapper that does the paging and exception handling as a cross-cutting concern.

stephenfuqua: Will rename `closure` and add types.

The `Facade` class's purpose is more than just paging. Name was definitely lazy and I'll work on a better one. Probably won't turn try to turn it into a decorator at this point, since it is performing business logic, not just cross-cutting concerns.

stephenfuqua: Done

bradbanister: @stephenfuqua  So, if going decorator is too much for now, you could extract the four "while True -> get_next_page() -> break" paging behaviors into a single method/function.

src/schoology-extractor/schoology_extractor/schoology_request_service.py
```diff
@@ -0,0 +1,112 @@
...
+
+        self._logger.debug("Exporting sections: get sections for active courses")
+
+        # TODO: inconsistency - in this case the pagination is handled inside of
+        # the client, instead of being handled here. Arguably there should be
+        # some other thing, neither this service nor the client, that handles
+        # the paging. As BB said in a pr comment, perhaps it should be handled
+        # in a decorator. This works but we should consider refactoring to a
+        # cleaner approach.
+        return self._client.get_section_by_course_ids([c["id"] for c in courses_list])
```
stephenfuqua, line 78: @lalrg  @bradbanister please note this TODO comment. I am going to create a stand-alone ticket to handle refactoring these existing methods. Luis, with your work next week you'll be adding new resources that need paging. The decorator idea is interesting but that might be a lot for any of us to take in this new language, especially without anyone else to talk through it with.

May I suggest that the paging for now should go into the `RequestClient`? I can see arguments either way. Ultimately, I think this domain service layer should be insulated from details about paging, and therefore the paging support needs to be pushed into the API client.

src/schoology-extractor/schoology_extractor/schoology_request_service.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def get_assignments(
+        self, sections: List[Dict[str, Any]], grading_periods: List[str]
+    ) -> list:
+
+        # NOTE: this call handles paging on its own
```
stephenfuqua, line 84: We can get rid of this comment in the future. I was just pointing out the inconsistency.

src/schoology-extractor/tests/test_schoology_request_service.py
```diff
@@ -0,0 +1,367 @@
...
+            )
+
+            roles = {"role": [{"id": 321, "title": "estudiante"}]}
+            roles_page = PaginatedResult(
+                request_client, page_size, roles, "role", "ignore me"
+            )
+
+            # Arrange
+            request_client.get_users.return_value = users_page
+            request_client.get_roles.return_value = roles_page
```
stephenfuqua, line 43: Here's where I was getting tripped up yesterday morning: I thought that I needed `request_client.return_value`. That would have been true if I had been using `mock.patch`, but in this case `request_client` _is_ the mock object already.

src/schoology-extractor/tests/test_schoology_request_service.py
```diff
@@ -0,0 +1,367 @@
...
+
+            # Act
+            result = service.get_users()
+
+            return result
+
+        def it_should_return_a_data_frame(result):
+            assert isinstance(result, pd.DataFrame)
+
+    def describe_given_two_pages_of_users():
```
stephenfuqua, line 59: Overall I liked using `describe`'s improvement to the syntax, and I like the use of `@pytest.fixture` to create a fixture that does all of the work, followed by individual tests that run small asserts.

The only weird thing was that I sometimes needed to access the mock object in the test function, and thus needed to return the mock from the fixture.

src/schoology-extractor/schoology_extractor/api/request_client.py
lalrg: I'd at least include those values in the `.env.example` or in the `readme` file.

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -34,19 +45,14 @@ class RequestClient:
...
-        assert isinstance(base_url, str), "Argument `base_url` should be of type `str`."
+    schoology_key: str
+    schoology_secret: str
+    base_url: str = DEFAULT_URL
 
-        self.oauth = OAuth1Session(schoology_key, schoology_secret)
-        self.base_url = base_url
-        self.consumer_key = schoology_key
-        self.consumer_secret = schoology_secret
+    def __post_init__(self):
```
lalrg, line 54: I'd move the asserts for the properties to this method... What I don't like about having them in the _request_header prop, is that if later we need to add another property or method, is not clear what values we need to validate or where, while if we have them here, it would be like a central point to evaluate all the required properties for the class that should be passed in the constructor.

src/schoology-extractor/schoology_extractor/api/request_client.py
stephenfuqua: The RETRY ones should be included and I'll add that to the readme. The base url I would rather keep hidden for now.

src/schoology-extractor/schoology_extractor/schoology_request_service.py
stephenfuqua: https://tracker.ed-fi.org/browse/FIZZ-142

src/schoology-extractor/pyproject.toml
```diff
@@ -25,6 +26,8 @@ pyfakefs = "^4.1.0"
 coverage = "^5.3"
 mypy = "^0.782"
 teamcity-messages = "^1.28"
+requests-mock = "^1.8.0"
```
stephenfuqua, line 29: @bradbanister this package made it easy for me to do some API testing.

src/schoology-extractor/schoology_extractor/api/request_client.py
stephenfuqua: Done. And more 😆.

src/schoology-extractor/README.md
```diff
@@ -1,71 +1,120 @@
...
+variables. Environment variables can be set the normal way, or by using a
+dedicated [`.env` file](https://pypi.org/project/python-dotenv/). For `.env`
+support, we provided a [.env.example](.env.example) which you can copy, rename
+to `.env`, and adjust to your desired parameters. Supported parameters:
+
+| Description | Required | Command Line Argument | Environment Variable |
+| ----------- | -------- | --------------------- | -------------------- |
+| Schoology API Key | yes | -k or --client-key | SCHOOLOGY_KEY |
+| Schoology API Secret | yes |  -s or --client-secret | SCHOOLOGY_SECRET |
+| List of Grading Periods | yes | -g or --grading-periods | SCHOOLOGY_GRADIN_PERIODS |
```
lalrg, line 28: Great Idea, just a small change
```suggestion
| List of Grading Periods | yes | -g or --grading-periods | SCHOOLOGY_GRADING_PERIODS |
```

src/schoology-extractor/schoology_extractor/api/request_client.py
lalrg: Makes sense to me

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -294,10 +343,8 @@ def get_roles(self, page_size: int = DEFAULT_PAGE_SIZE) -> PaginatedResult:
...
 
-        return PaginatedResult(
-            self, page_size, response, "role", self.base_url + url
-        )
+        return PaginatedResult(self, page_size, response, "role", self.base_url + url)
 
     def _build_query_params_for_first_page(self, page_size: int):
-        assert isinstance(page_size, int)
+        assert isinstance(page_size, int), "Argument `page_size` should be of type `int`."
         return f"start=0&limit={page_size}"
```
lalrg, line 350: I think this file is starting to get too big for a single file, if it keeps growing, sooner than later it will start getting difficult to read/understand.

src/schoology-extractor/schoology_extractor/grading_periods.py
```diff
@@ -25,21 +25,26 @@
...
 logger.info("Ed-Fi LMS Schoology Extractor - Grading Periods")
 request_client = RequestClient(schoology_key, schoology_secret)
 
 logger.info("Getting grading periods")
+
 try:
+
     grading_periods_list: List[Any] = []
     grading_periods_response = request_client.get_grading_periods()
     while True:
```
lalrg, line 41: Try to execute as few lines as possible in try/catch blocks. Initializing a variable should not fail.
```suggestion
    grading_periods_list: List[Any] = []

try:

    grading_periods_response = request_client.get_grading_periods()
    while True:
```

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -27,115 +27,82 @@
...
-
-    export_data.to_csv(
-        submissions_list, os.path.join(schoology_output_path, "submissions.csv")
-    )
-except Exception as ex:
-    logger.error('An exception has occurred in the process of generating the submissions.csv file: %s', ex)
+
+
+def _get_users():
+    return service.get_users()
```
lalrg, line 75: Is this function necessary?

src/schoology-extractor/schoology_extractor/schoology_request_service.py
```diff
@@ -0,0 +1,112 @@
+# SPDX-License-Identifier: Apache-2.0
```
lalrg, line 1: I am not sure if schoology_request_service is a good name for this file... Visitors might infer from the name, that they should look into this file for making requests to the Schoology API, while this is actually both getting data from the API(but not directly exposing the get and other request methods inside of request_client) and exporting it to csv, to then be called by the main script. 
I would also add docstrings to this class.

src/schoology-extractor/tests/api/test_request_client.py
```diff
@@ -32,247 +31,276 @@ def test_given_custom_url_then_use_it_for_the_base_url(self):
...
 
         def test_then_oauth_word_is_present(self, default_request_client):
             assert self._present_in_header(
                 default_request_client._request_header,
                 header_key="Authorization",
-                contained_value="OAuth"
+                contained_value="OAuth",
             )
 
         def _present_in_header(self, request_header, header_key, contained_value):
```
lalrg, line 54: I'd put this kind of functions outside of the class.

stephenfuqua: You mean this "private" method? If it isn't needed anywhere else, why put it outside the class?

src/schoology-extractor/schoology_extractor/schoology_request_service.py
stephenfuqua: I don't particularly like the name but have not been able to come up with anything better. I have not found a good, consistent pattern name that applies to the main business logic of a back-end component. At first I thought of facade:

> "Provide a unified interface to a set of interfaces in a subsystem. Facade defines a higher-level interface that makes the subsystem easier to use."

That is close, but in my experience a facade usually sits over one service or a couple of very similar services that belong together. In this case we are orchestrating the calls to the API service and to the filesystem.

Orchestrator is a possibility. However, I usually think of the `main` script more with that term. 

I used the word "service" when about this being a _service layer_. Maybe `schoology_extractor_service` would be better?

src/schoology-extractor/schoology_extractor/main.py
stephenfuqua: No, it is not necessary. I only created it because it felt weird to have all of the other calls wrapped in a function but not this one, like this:

```python
_create_file_from_dataframe(service.get_users(), "users.csv")
_create_file_from_list(_get_sections, "sections.csv")
_create_file_from_list(_get_assignments, "assignments.csv")
_create_file_from_list(_get_submissions, "submissions.csv")
```

That just looks so strange to me that I preferred giving it a separate function. But I'm not attached to it.

src/schoology-extractor/schoology_extractor/grading_periods.py
stephenfuqua: While I understand the sentiment, there is no functional difference so I don't see any value to changing the code here.

src/schoology-extractor/schoology_extractor/api/request_client.py
stephenfuqua: Many of the lines are just from docstrings 😆 .

The paging code can be moved to a helper module.  Perhaps `get_assignments_by_section_ids` should be changed to move the looping over sections out to the `schoology_request_service` (or whatever name it gets) so that the `request_client` module is more low-level, without that business logic. There's an opportunity to refactor this in the separate ticket I created for refactoring the paging support.

src/schoology-extractor/README.md
stephenfuqua: Done

src/schoology-extractor/tests/api/test_request_client.py
lalrg: I was thinking about leaving it outside of the class because of it does not depend on the class nor helps with readability for the test, but yes, if it is only used here, makes sense to have it here.

src/schoology-extractor/schoology_extractor/schoology_request_service.py
lalrg: Yeah, schoology_extractor_service sounds better, but I'd add docstring to that class

src/schoology-extractor/schoology_extractor/main.py
lalrg: Okay, it is fine, I like the idea of consistent calls in the main script, improves readability.

src/schoology-extractor/schoology_extractor/schoology_request_service.py
stephenfuqua: After looking at this more closely again, I realized I was getting myself confused: it is not exporting to CSV. It is just getting data and reshaping. So "extractor" is not good in the name. "Facade" comes back into favor then, since a Facade will often reshape a Data Transfer Object to a Domain Object, just as we're doing now. So I'm changing to `schoology_extract_facade`. Note "extract" instead of "extractor"

lalrg (approved):
I reviewed the code and tested the functionality and it is working as expected.

PR #57, opened 2020-10-16
[FIZZ-102]

stephenfuqua:
* Maps the Schoology Assignments format to the Ed-Fi LMS UDM format
* Writes assignments to the correct directory, using datetime for the file name
* Opportunistically fix an oversight for the Users file: write to the correct directory, using datetime for the file name


src/schoology-extractor/schoology_extractor/mapping/assignments.py
```diff
@@ -0,0 +1,79 @@
...
+    ----------
+    assignments_df: DataFrame
+        Pandas DataFrame containing Schoology assignments for a section
+
+    Returns
+    -------
+    DataFrame
+        A LMSUsers-formatted DataFrame
+
+    DataFrame columns are:
```
lalrg, line 28: Add this to Notes section.

PR #58, opened 2020-10-18
[FIZZ-108] generate user-section associations

bradbanister:


bradbanister:
User-Section Association generation, the main point of the story, is complete so please review.  I am still working on general API endpoint mocking/testing, which is why this is a draft.

src/google-classroom-extractor/google_classroom_extractor/mapping/user_section_associations.py
```diff
@@ -0,0 +1,167 @@
...
+        StartDate: Month, day, and year of the user's entry or assignment to the section
+        LMSUserIdentifier: A unique numeric identifier assigned to the user
+        LMSUserLMSSectionAssociationIdentifier: A unique numeric identifier assigned to
+            the user section association
+        CreateDate: Date this record was created
+        LastModifiedDate: Date this record was last updated
+    """
+    assert isinstance(students_df, DataFrame)
+    assert "userId" in students_df.columns
+    assert "courseId" in students_df.columns
```
lalrg, line 134: Since this assert is being called in different places, you could move the `"userId"` and `"courseId"` validations to an external method and call it `_has_required_columns` or something like that

src/google-classroom-extractor/tests/api/api_caller/test_api_caller.py
```diff
@@ -0,0 +1,20 @@
+# SPDX-License-Identifier: Apache-2.0
```
lalrg, line 1: Why is this file commented?

src/google-classroom-extractor/google_classroom_extractor/mapping/user_section_associations.py
bradbanister: Unfortunately, asserts are documentation as much as anything else so I don't think it makes sense to try to extract common assertions.  Another way to think about it is that asserts can be disabled in production.  That would lead to an extracted function with no behavior.

src/google-classroom-extractor/tests/api/api_caller/test_api_caller.py
bradbanister: Removed - it's part of the API testing i was working on.

lalrg (approved):
Looks good to me

PR #59, opened 2020-10-21
[FIZZ-102] Changes from code review

lalrg:
It was originally a change implemented by Stephen, I merged by accident so I created this PR to implement some changes from code review.
For testing it, you should set up your environment, following the instructions in the [readme file](https://github.com/Ed-Fi-Exchange-OSS/Ed-Fi-X-Fizz/blob/FIZZ-102-code-review/src/schoology-extractor/README.md) for the extractor. I can get you the .env file on slack in case you need it, just let me know if it is the case.

The test should consist on two steps:
* Making sure that by running `poetry run python .\schoology_extractor\main.py` you get the expected output according with the [AC for this story](https://tracker.ed-fi.org/browse/FIZZ-102).
* Making sure that by running `poetry run pytest`, all the test are passing.

After this, the PR can be merged into october branch.

### DIFF
--- a/src/schoology-extractor/schoology_extractor/main.py
+++ b/src/schoology-extractor/schoology_extractor/main.py
@@ -51,8 +51,9 @@ service = SchoologyExtractFacade(logger, request_client, page_size)
 def _create_file_from_dataframe(action: Callable, file_name):
     logger.info(f"Exporting {file_name}")
     try:
-        data = action()
-        export_data.df_to_csv(data, file_name)
+        data: pd.DataFrame = action()
+        if data is not None:
+            export_data.df_to_csv(data, file_name)
     except Exception as ex:
         logger.error(
             f"An exception occurred while generating {file_name} : %s",
@@ -66,7 +67,8 @@ def _create_file_from_list(action: Callable, file_name):
     logger.info(f"Exporting {file_name}")
     try:
         data = action()
-        export_data.to_csv(data, os.path.join(schoology_output_path, file_name))
+        if data is not None:
+            export_data.to_csv(data, os.path.join(schoology_output_path, file_name))
     except Exception as ex:
         logger.error(
             f"An exception occurred while generating {file_name} : %s",
@@ -93,7 +95,7 @@ def _get_assignments(section_id) -> Callable:
     def __get_assignments():
         assignments = service.get_assignments(section_id)
         result_bucket["assignments"] = assignments
-        return assignments
+        return service.map_assignments_to_udm(assignments, section_id) if not assignments.empty else None
 
     return __get_assignments
 
--- a/src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
+++ b/src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
@@ -5,7 +5,7 @@
 
 from dataclasses import dataclass
 from logging import Logger
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Union
 
 import pandas as pd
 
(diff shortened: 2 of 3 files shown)


src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -51,8 +51,9 @@
 def _create_file_from_dataframe(action: Callable, file_name):
     logger.info(f"Exporting {file_name}")
     try:
-        data = action()
-        export_data.df_to_csv(data, file_name)
+        data: pd.DataFrame = action()
+        if data is not None:
+            export_data.df_to_csv(data, file_name)
```
stephenfuqua, line 56: 👍

src/schoology-extractor/tests/test_schoology_extract_facade.py
```diff
@@ -242,8 +242,9 @@ def system() -> Tuple[pd.DataFrame, Mock]:
 
             # Act
             result = service.get_assignments(section_id)
+            mapped_result = service.map_assignments_to_udm(result, section_id)
```
stephenfuqua, line 245: Can you separate out the unit tests, so that the test is only covering one function instead of both `get_assignments` and `map_assignments_to_udm`?

src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
```diff
@@ -113,12 +113,11 @@ def get_assignments(self, section_id: int) -> pd.DataFrame:
...
+        return pd.DataFrame(self._client.get_assignments(section_id, self._page_size))
 
-        assignments = self._client.get_assignments(section_id, self._page_size)
-        df = pd.DataFrame(assignments)
-        df["section_id"] = section_id
-
-        return assignmentsMap.map_to_udm(df)
+    def map_assignments_to_udm(self, assignments: pd.DataFrame, section_id: Union[int, str]):
+        assignments["section_id"] = section_id
+        return assignmentsMap.map_to_udm(assignments)
```
stephenfuqua, line 120: I don't see the value in splitting this up.

src/schoology-extractor/tests/test_schoology_extract_facade.py
lalrg: I will.

src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
lalrg: get_submissions method from facade, receives the assignments in their original shape, so I hadto get the original data, assign it to the result_bucket in main, and then map it to udm.

```python3
assignments = service.get_assignments(section_id)
result_bucket["assignments"] = assignments
return service.map_assignments_to_udm(assignments, section_id) if not assignments.empty else None
```

lalrg: And I could not use the udm version, because get_submissions needs access to grade_item_id, which  is not part of UDM version, it is part of the original API response.

src/schoology-extractor/tests/test_schoology_extract_facade.py
lalrg: ✅

src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
stephenfuqua: OH, I see now.

stephenfuqua (approved):

PR #60, opened 2020-10-21
[FIZZ-124] Synchronization process and Create / Last Modified Dates for the users file

lalrg:
For testing, you should set up your environment, following the instructions in the [readme file](https://github.com/Ed-Fi-Exchange-OSS/Ed-Fi-X-Fizz/blob/FIZZ-124/src/schoology-extractor/README.md) for the extractor. I can get you the .env file on slack in case you need it, just let me know if it is the case.

After the review, the PR can be merged into main branch.

### DIFF
--- a/src/schoology-extractor/data/.gitignore
+++ b/src/schoology-extractor/data/.gitignore
@@ -1,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file - it's the directory placeholder
+!.gitignore
--- a/src/schoology-extractor/schoology_extractor/api/request_client.py
+++ b/src/schoology-extractor/schoology_extractor/api/request_client.py
@@ -15,7 +15,7 @@ from requests.packages.urllib3.exceptions import ProtocolError  # type: ignore
 from requests_oauthlib import OAuth1Session  # type: ignore
 
 from .paginated_result import PaginatedResult
-
+from schoology_extractor.helpers.constants import RESOURCE_NAMES
 
 DEFAULT_URL = os.environ.get("SCHOOLOGY_BASE_URL") or "https://api.schoology.com/v1/"
 DEFAULT_PAGE_SIZE = 20
@@ -164,7 +164,7 @@ class RequestClient:
                 self,
                 page_size,
                 self.get(url),
-                "assignment",
+                RESOURCE_NAMES.ASSIGNMENT,
                 self.base_url + url,
             )
             while True:
@@ -246,7 +246,10 @@ class RequestClient:
         for course_id in course_ids:
             url = f"courses/{course_id}/sections"
             sections_per_course = PaginatedResult(
-                self, page_size, self.get(url), "section", self.base_url + url
+                self, page_size,
+                self.get(url),
+                RESOURCE_NAMES.SECTION,
+                self.base_url + url
             )
             while True:
                 sections = sections + sections_per_course.current_page_items
@@ -289,7 +292,11 @@ class RequestClient:
 
         response = self.get(url)
         return PaginatedResult(
-            self, page_size, response, "revision", self.base_url + url
+            self,
+            page_size,
+            response,
+            RESOURCE_NAMES.REVISION,
+            self.base_url + url
         )
 
     def get_users(self, page_size: int = DEFAULT_PAGE_SIZE) -> PaginatedResult:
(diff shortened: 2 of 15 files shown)


ejansson:
@lalrg looks like the CI build failed - is this expected?

lalrg:
> @lalrg looks like the CI build failed - is this expected?

@ejansson Yes, it is expected. I am currently working on Unit Testing the new functionality, and it is failing when it runs the Unit Tests, so it is ok.

src/schoology-extractor/data/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file - it's the directory placeholder
+!.gitignore
```
stephenfuqua, line 4: Standard convention is to have a `.gitkeep` file in this directory. The root level `.gitignore` should already be ignoring `*.csv`.

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -164,7 +164,7 @@ def get_assignments_by_section_ids(
                 self,
                 page_size,
                 self.get(url),
-                "assignment",
+                RESOURCE_NAMES.ASSIGNMENT,
```
stephenfuqua, line 167: Generally I have pulled strings like this into a constant if using in multiple places and change could happen. Makes me wonder if the UDM mappers should have more column names as constants. Worth thinking about - we might want to go ahead and change those mappers to use constants in more places.

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -0,0 +1,116 @@
...
+    with db_engine.connect() as con:
+        result: Union[ResultProxy, None] = con.execute(
+            f"SELECT CreateDate from {resource_name} "
+            f"WHERE {unique_id_column} == {id}"
+        )
+        create_date = result.first()
+    return create_date[0] if create_date is not None else _get_current_date_with_format()
+
+
+def sync_resource(
```
stephenfuqua, line 95: Should this one have a docstring, since it is a "public" method?

src/schoology-extractor/data/.gitignore
stephenfuqua: But this file is fine too I suppose 😃 .

stephenfuqua (changes requested):
Very small change request. Would like for @bradbanister to take a look at this one too due to familiarity with working with Sqlite.

src/schoology-extractor/data/.gitignore
lalrg: I made the same comment in a past PR from Brad. I think it is fine to have a .gitignore ignoring all the contained files, because of I think we should not commit anything contained in this directory.

bradbanister: I really like this style as opposed to the .gitkeep convention.  The ignore everything prevents someone from ever accidentally placing a new (non-ignored elsewhere) file type in there and accidentally committing.  Also, it doesn't rely on convention.

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -0,0 +1,116 @@
...
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+from datetime import datetime
+import os
+import logging
+from typing import Optional, Union
+
+from pandas import DataFrame
+import sqlalchemy
+from sqlalchemy.engine import ResultProxy
```
bradbanister, line 12: cool that you found this type

bradbanister (approved):
Sync process looks good

src/schoology-extractor/data/.gitignore
stephenfuqua: Good case for this approach instead of `.gitkeep`.

stephenfuqua (changes requested):
QA testing: need to set `LastModifiedDate` only if the record has changes.

src/schoology-extractor/tests/mapping/test_assignments.py
```diff
@@ -26,7 +26,7 @@ def result() -> pd.DataFrame:
             "title": "The title",
             "type": "assignment",
             "description": "The description",
-            "due": "8/21/2020  11:59:00 PM",
+            "due": "2020-8-21  23:59:00",
```
stephenfuqua, line 29: The actual API results have a single space between the date and time. Does that matter? Will it be ok for this test value to have two spaces?

lalrg: fixed

stephenfuqua (approved):

PR #66, opened 2020-10-28
FIZZ-125 Schoology datetime stamp for sections file

lalrg:


src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -40,29 +41,41 @@ def _resource_has_changed(
     db_engine:  sqlalchemy.engine.base.Engine,
     id_column: str = "id"
 ) -> bool:
-    query = f"SELECT * FROM {resource_name} where {id_column} = {resource[id_column]} limit 1;"
+    query = f"SELECT * FROM {resource_name} where SourceId = {resource[id_column]} limit 1;"
```
lalrg, line 44: `SourceId` will be the identifier from the LMS, it will be exposed in a column to simplify the process of querying existing resources to check if they have been updated.

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -40,29 +41,41 @@ def _resource_has_changed(
...
         if result is None:
             return True
         for row in result:
-            for column, value in row.items():
-                if column != "CreateDate" and column != "LastModifiedDate":
-                    if resource[column] != value:
-                        return True
+            for column, db_value in row.items():
+                db_item = {**db_item, column: db_value}
+    json_response = r"%s" % db_item["JsonResponse"]
```
lalrg, line 53: `r` is for raw string, without it, `json.loads` can fail.

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -169,11 +177,20 @@ def sync_resource(
...
+            JsonResponse=json.dumps(row)
+        )
+        data_to_write.append(record)
+        row["CreateDate"] = create_date
+        row["LastModifiedDate"] = last_modified_date
 
-    df_data = DataFrame(data)
-    _write_resource_to_db(resource_name, db_engine, df_data, column_mapping)
-    _remove_duplicates(resource_name, db_engine, id_column)
+    df_data = DataFrame(data_to_write)
```
lalrg, line 192: 2 Different formats are needed:
1. Used for syncDB
```python3 
record = dict(
            CreateDate=create_date,
            LastModifiedDate=last_modified_date,
            SourceId=row[id_column],
            JsonResponse=json.dumps(row)
        )
```

2. Used for mapping and exporting to csv.
```python3
# row is the original item, we add this properties to it with the help of the sync process
row["CreateDate"] = create_date
row["LastModifiedDate"] = last_modified_date
```

For that reason, we need two dataframes: one as output of the process, another one to write to sync_db.

src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
```diff
@@ -92,8 +91,8 @@ def get_users(self) -> pd.DataFrame:
         roles_df = sync.sync_resource(
             RESOURCE_NAMES.ROLE,
             self._db_engine,
-            roles_list,
-            sync_column_types.ROLE_COLUMN_TYPES_MAPPING)
+            roles_list
+            )
```
lalrg, line 95: Now we don't need to map the column types, everything is written as a string in a single column.

src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
```diff
@@ -122,9 +121,11 @@ def get_sections(self) -> pd.DataFrame:
         # the paging. As BB said in a pr comment, perhaps it should be handled
         # in a decorator. This works but we should consider refactoring to a
         # cleaner approach.
-        sections = pd.DataFrame(
+        sections = sync.sync_resource(
+            RESOURCE_NAMES.SECTION,
+            self._db_engine,
             self._client.get_section_by_course_ids([c["id"] for c in courses_list])
-        )
+            )
```
lalrg, line 128: The implementation of sync_resource for further resources, should be as simple as it is here.

src/schoology-extractor/tests/helpers/test_sync.py
```diff
@@ -169,19 +165,20 @@ def test_return_current_date_if_there_is_no_record(self):
...
-    # [(key0, value0), (key1, value1)]
     row = Mock()
     row.items.return_value = [
-        ('id', 1),
-        ('name', 'fake_name')
+        ('JsonResponse',
+            r"""{
+                "id": 1,
+                "name": "fake_name"
+            }""")
```
lalrg, line 181: `JsonResponse` is only one of the returned keys, for the scope of this test, this key is the only required one.

src/schoology-extractor/schoology_extractor/helpers/sync.py
stephenfuqua: In at least one case - attendance events - the SourceId will be a string instead of an integer. This is because Schoology does not assign a source id for attendance. I created one with `f"{enrollment_id}#{event_date}`. I assume we will need to change the syntax here to have `... SourceId = '{resource[id_column]}' ... ` when we get to the story for setting sync dates on attendance events. Can you check if this will be a problem for integer values? I'm not asking to change it right now, as it is not necessary right now.

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -40,29 +41,38 @@ def _resource_has_changed(
...
         result: Union[ResultProxy, None] = con.execute(query)
         if result is None:
             return True
         for row in result:
-            for column, value in row.items():
-                if column != "CreateDate" and column != "LastModifiedDate":
-                    if resource[column] != value:
-                        return True
+            for column, db_value in row.items():
+                db_item = {**db_item, column: db_value}
```
stephenfuqua, line 52: Why not just use `db_item[column] = db_value` here?

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -40,29 +41,38 @@ def _resource_has_changed(
...
+    json_response = r"%s" % db_item["JsonResponse"]
+    db_parsed_item = json.loads(json_response)
+
+    for key in db_parsed_item:
+        api_value = resource[key]
+        db_value = db_parsed_item[key]
+
+        if isinstance(api_value, dict) or isinstance(api_value, list):
+            api_value = str(api_value).replace("'", '"')
+            db_value = str(db_value).replace("'", '"')
```
stephenfuqua, line 62: Is this conversion still necessary?

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -85,28 +95,26 @@ def _table_exist(
...
-    unique_id_column: str
+    db_engine: sqlalchemy.engine.base.Engine
 ):
     with db_engine.connect() as con:
         con.execute(
             f"DELETE from {resource_name} "
             "WHERE rowid not in (select max(rowid) "
             f"FROM {resource_name} "
-            f"GROUP BY {unique_id_column})"
+            f"GROUP BY SourceId)"
```
stephenfuqua, line 105: I have never seen a `DELETE` statement with a `GROUP BY` statement before! Nice technique.

stephenfuqua (commented):
I have a few questions, nothing major. Going to run QA tests and give you a chance to think about questions before approving.

src/schoology-extractor/schoology_extractor/helpers/sync.py
bradbanister: Good to know

src/schoology-extractor/schoology_extractor/helpers/sync.py
bradbanister: Google Classroom does this too.  😎

bradbanister (commented):
This looks good to me.  My comments are only commentary.  I'll let Stephen give final approval though, since he's made a pass through already and knows this code better.

src/schoology-extractor/schoology_extractor/helpers/sync.py
lalrg: No reason. It could be changed.

src/schoology-extractor/schoology_extractor/helpers/sync.py
lalrg: Yes, for complex types it is, it could be improved, but it is necessary.

src/schoology-extractor/schoology_extractor/helpers/sync.py
lalrg: Got it from Brad's code 😄

src/schoology-extractor/schoology_extractor/helpers/sync.py
lalrg: I could define one argument for a "custom key", and use it when it is not null, so we could use it for specific resources.

lalrg: I made the change to set all SourceId columns to string(varchar, probably) in the sync database.

stephenfuqua (approved):

stephenfuqua (changes requested):
`sections.py` needs to be updated - it is still hard coding the date columns to `None`.

stephenfuqua (approved):

PR #67, opened 2020-10-30
[FIZZ-106] Schoology attendance events export

stephenfuqua:


src/schoology-extractor/schoology_extractor/mapping/attendance.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+
+            df = df.append(reshape)
+
+    return df.convert_dtypes()
+
+
+def __get_status(status_code: int) -> str:
+    switcher = {1: "present", 2: "absent", 3: "late", 4: "excused"}
+    return switcher.get(status_code, f"Unknown status: {status_code}")
```
stephenfuqua, line 40: I have used underscore inconsistently here: just a single underscore for the first method, and double underscore for the second. Why did it happen? Indecisiveness. In both cases I have no need to directly unit test the function. They are both small local/private helper functions that keep `map_to_udm` from becoming too large. @lalrg @bradbanister do you have a preference between single and double underscore for functions like these? I've read about the difference before. At this point it feels more like a style decision to me, unless for some reason the "private" function is going to be mocked - then single underscore might be preferable?

(Why would you mock a helper function? Because that helper function is "hiding" access to something much more difficult to mock, such as a small routine for database access).

src/schoology-extractor/schoology_extractor/mapping/attendance.py
```diff
@@ -0,0 +1,121 @@
...
+        Date: Attendance date, formatted as YYYY-mm-dd
+        AttendanceStatus: one of (present, absent, late, excused)
+        LMSUserSourceSystemIdentifier: source system identifier for the user
+        LMSUserLMSSectionAssociationSourceSystemIdentifier: source system identifier for the section association
+        EntityStatus: The status of the record
+        CreateDate: datetime at which the record was first retrieved
+        LastModifieDate: datetime when the record was modified, or when first retrieved
+    """
+    if len(attendance) == 0:
+        return pd.DataFrame()
```
stephenfuqua, line 74: This kind of check is not built into the other mapping functions, although perhaps it should be. In this case there are some sections for which no attendance has been recorded, leading to a completely empty initial list.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
```diff
@@ -0,0 +1,121 @@
...
+        [
+            "SourceSystemIdentifier",
+            "LMSUserSourceSystemIdentifier",
+            "LMSSectionSourceSystemIdentifier",
+        ]
+    ].copy()
+    # This data type conversion was required because Schoology is returning
+    # enrollment Id as an integer in the Attendance endpoint, but as a string
+    # with the Enrollment endpoint.
+    sa["SourceSystemIdentifier"] = sa["SourceSystemIdentifier"].apply(int)
```
stephenfuqua, line 96: This drove me crazy. The merge statement below was working in unit testing, but failing in full application testing. Finally I realized that the failure was due to a type mismatch.

There is at least one more way to set a data type for a column: [astype()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.astype.html). They seem equally reasonable to me, and the approach above is the first one I learned so it was my default.

src/schoology-extractor/schoology_extractor/mapping/section_associations.py
```diff
@@ -43,7 +43,7 @@ def map_to_udm(enrollments_df: pd.DataFrame, section_id: int) -> pd.DataFrame:
         EnrollmentStatus: Possible values are Active, Expired, Invite pending, Request Pending, Archived
         StartDate: Date when the section-association becomes Active
         EndDate: Date when the section-association changes from Active to another status
-        UserSourceSystemIdentifier: A unique number or alphanumeric code assigned to a user by the source
+        LMSUserSourceSystemIdentifier: A unique number or alphanumeric code assigned to a user by the source
```
stephenfuqua, line 46: In the current story I realized that this field was misnamed in a prior story.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
bradbanister: Funny, I was thinking about this situation yesterday.  I'm fine with single underscore in your double underscore case.  I'm a fan of method extraction, often just to give a name to a particular operation being performed.  I think it's ok not to try to differentiate the reason a function is private.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
bradbanister: I like this. There are probably degenerate cases like this in GC that I don't know about.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
bradbanister: I only wonder if astype() would avoid the iteration across all rows that the apply() requires.

bradbanister (changes requested):
The .mypy.ini file appears to be mis-named ".mypi.ini", causing mypy failures.  Also, the README section on output needs updating.  Everything else looks good, other comments are suggestions.

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -112,11 +115,24 @@ def _get_submissions() -> list:
 
 def _get_section_associations(section_id: int) -> Callable:
     def __get_section_associations() -> pd.DataFrame:
-        return service.get_section_associations(section_id)
+        section_associations = service.get_section_associations(section_id)
+        result_bucket["section_associations"] = section_associations
```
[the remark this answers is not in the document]
stephenfuqua: I'll look into that.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
lalrg: I like single underscore.

stephenfuqua:
> The .mypy.ini file appears to be mis-named ".mypi.ini", causing mypy failures. Also, the README section on output needs updating. Everything else looks good, other comments are suggestions.

Do'h! It works on the build server because the build.py is configured to look for `mypi.ini` with the misspelling. I'll fix in both places. Here I was thinking that `poetry run mypy ./schoology_extractor` was failing because I was being lazy and not providing the argument for the config file, without realizing that it would pick up the config file if it were named correctly.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
stephenfuqua: Since I'm going in to make a few other changes, I'm going to update this.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
stephenfuqua: Going to change to single underscore.

stephenfuqua: Done

src/schoology-extractor/schoology_extractor/mapping/attendance.py
stephenfuqua: Done

stephenfuqua:
@bradbanister changes are in

bradbanister (approved):
Latest updates look good

PR #68, opened 2020-10-30
FIZZ-110: Google Classroom: User-Activities Data Output Standardization

bradbanister:
Tests are forthcoming


src/google-classroom-extractor/google_classroom_extractor/csv_generation/write.py
```diff
@@ -14,6 +14,7 @@
 SECTION_ASSOCIATIONS_ROOT_DIRECTORY = "data/ed-fi-udm-lms/section={id}/section-associations/"
 ASSIGNMENT_ROOT_DIRECTORY = "data/ed-fi-udm-lms/section={id}/assignments/"
 SUBMISSION_ROOT_DIRECTORY = "data/ed-fi-udm-lms/section={id1}/assignment={id2}/submissions/"
+USER_ACTIVITY_ROOT_DIRECTORY = "data/ed-fi-udm-lms/section={id}/user-activities/"
```
stephenfuqua, line 17: In the Schoology extractor, we're note hard-coding `data/ed-fi-udm-lms` - instead handling this "base directory" as configuration parameter. Let's leave it alone for now, and perhaps we can change this in [FIZZ-62](https://tracker.ed-fi.org/browse/FIZZ-62)?

stephenfuqua: In fact, I have a module just for determining the file paths. Perhaps that module should be extracted to a package that can be shared with all of the extractors.

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
```diff
@@ -0,0 +1,154 @@
...
+        LastModifiedDate: Date this record was last updated
+    """
+    assert isinstance(submissions_df, DataFrame)
+    assert "submissionHistory" in submissions_df.columns
+    assert "id" in submissions_df.columns
+    assert "courseId" in submissions_df.columns
+    assert "courseWorkId" in submissions_df.columns
+
+    # convert json-like submissionHistory string to list of dicts
+    submissions_df["submissionHistory"] = submissions_df["submissionHistory"].apply(lambda json_like: json.loads(json_like.replace("'", '"')))
```
stephenfuqua, line 58: @lalrg I see that Brad is also doing a swap of `'` and `"` for consistency.

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
```diff
@@ -0,0 +1,154 @@
...
+    history_df.drop(columns=["submissionHistory"], inplace=True)
+
+    # expand stateHistory
+    state_history_df = concat([history_df, history_df["stateHistory"].apply(Series)], axis=1)
+    state_history_df.dropna(subset=["stateHistory"], inplace=True)
+
+    # enrich state_history_df
+    state_history_df["SourceSystemIdentifier"] = "S-" + state_history_df[
+        ["courseId", "courseWorkId", "id", "stateTimestamp"]
+    ].agg("-".join, axis=1)
```
stephenfuqua, line 76: In Schoology attendance events there is no source system identifier, so I had to create one by concatenating the enrollment id and the event date, giving me something like `123456-2020-10-30`. Then I found it a little odd for dashes to have two different meanings, and switched to `123456#2020-10-30`. Ultimately internal self-consistency in the application is all that matters. Thoughts?

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
```diff
@@ -0,0 +1,154 @@
...
+            "id": "AssignmentIdentifier",
+            "actorUserId": "LMSUserIdentifier",
+        }
+    )
+
+    state_history_df["LMSUserActivityIdentifier"] = state_history_df["SourceSystemIdentifier"]
+    state_history_df["ActivityType"] = "Submission State Change"
+    state_history_df["ActivityTimeInMinutes"] = ""
+    state_history_df["Content"] = ""
+    state_history_df["EntityStatus"] = ENTITY_STATUS_ACTIVE
```
stephenfuqua, line 103: I have yet to figure out the purpose of this field. I keep meaning to ask Gabrielle again. Thus there is never a case in Schoology where we're setting a value other than `active`. And I see that here with the user submissions. Are there any cases so far, in Google Classroom, where you have a value other than `active`?

src/google-classroom-extractor/google_classroom_extractor/csv_generation/write.py
bradbanister: That sounds like a good idea.

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
bradbanister: Yeah, this was the first time I had to include timestamps to differentiate. Hadn't thought about the double use of dashes. Not too concerned but open to changing. Hash is a good choice if we want to change as submission ids in GC can have underscores.

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
bradbanister: It's always Active so far.

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
bradbanister: Yeah, json.loads() complains otherwise. Doing exec() was another option, but no way am I going to execute something pulled from an API 😱

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
stephenfuqua: Leaving this to your judgment for now. Maybe a good thing to ask for feedback on from end-users ... assuming we ever have end users 😆 . 

I believe in the users

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
```diff
@@ -0,0 +1,153 @@
...
+        grade_history_df["ActivityType"] = ACTIVITY_TYPE_GRADE
+
+        # combine with stateHistory
+        user_submission_df = user_submission_df.append(grade_history_df)
+
+    # teacher actions can show up on student histories and vice-versa
+    user_submission_df.drop_duplicates(subset=["SourceSystemIdentifier"], inplace=True)
+
+    # finish with common columns
+    user_submission_df["LMSUserActivityIdentifier"] = user_submission_df["SourceSystemIdentifier"]
```
stephenfuqua, line 142: What is the purpose of this duplicated column? I didn't catch in the initial review, only when looking at the actual output. I did not expect this column, `LMSUserActivityIdentifier`.

Side note: the sample user activities file was not kept up to date with actual UDM changes and thus has invalid columns. I can update it in my current ticket.

stephenfuqua: Looking closely at the [UDM diagram](https://techdocs.ed-fi.org/display/EDFIODS/Fizz+-+LMS+Unified+Data+Model) again, I see that the `LMSUserActivity` table is described as having a primary key column `UserLMSActivityIdentifier`. Of course there's a small typo, as it should be `LMSUserActivityIdentifier` as you have it above. But I think this column would only show up in the RDBMS, as an auto-incrementing integer.

src/google-classroom-extractor/tests/api/courses/test_courses_request.py
```diff
@@ -0,0 +1,18 @@
...
+
+def describe_when_requesting_latest_courses():
+    @patch("google_classroom_extractor.api.courses.call_api")
+    def it_should_use_courses_list_resource(mock_call_api):
+        # arrange
+        resource = Mock()
+        # act
+        request_latest_courses_as_df(resource)
+        # assert
+        assert isinstance(resource.courses.return_value.list, Mock)
```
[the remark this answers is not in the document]
bradbanister: My original thought was that since I had sync integration tests in this area, I would do pure unit tests for the rest since the code was simple. Unfortunately, these functions are so small that unit test granularity lead me to a low-quality tautological test that I'm calling `call_api()` with the correct resource.  The only other thing to test at this level of granularity is that I'm using `json_normalize()`, another low-quality test.

src/google-classroom-extractor/google_classroom_extractor/mapping/user_submission_activities.py
bradbanister: I'll drop this.

src/google-classroom-extractor/tests/api/courses/test_courses_request.py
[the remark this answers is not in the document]
stephenfuqua: OK. I do see your point about the actual work in the function being very trivial - thanks to the great features available to us, like `json_normalize()`. But it could have been more complicated. What if there was something like a mock response with `{'id':1}` and validation that the return is a DataFrame containing that single entry?

stephenfuqua (approved):

PR #70, opened 2020-11-03
[FIZZ-145] Tutorial on Using the Filesystem and Analyzing Results

stephenfuqua:


stephenfuqua:
I should change the diagram and the sample files to use directory `user-activities` instead of directory `activities`.

ggaronzik (approved):

PR #74, opened 2020-11-05
[FIZZ-160] include LMSSectionIdentifier in UDM assignment output

bradbanister:
Ticket says name should be "LMSSectionSourceSystemIdentifier", but techdocs says "LMSSectionIdentifier".  Going with techdocs to start.


stephenfuqua:
> Ticket says name should be "LMSSectionSourceSystemIdentifier", but techdocs says "LMSSectionIdentifier". Going with techdocs to start.

I see that now, but I had never seen it before. Somehow that part table of attributes has escaped my attention. All of the code in Schoology and in my sample CSV files have used `SourceSystem` in the foreign key names. Let's talk about in standup.

stephenfuqua (approved):

PR #75, opened 2020-11-05
[FIZZ-126] Sync process for section associations

stephenfuqua:


lalrg (commented):
I don't know if this is part of the scope of the story, but I found something interesting.
I am not getting the expected output, the dates are set as expected, but not all data is being exported:
I only get a csv for one section 
image

The csv contains 5 rows of data:
image

In the DB I have 9 rows of data for at least two sections:
image

So, the sync process is working, but some data is not being exported to csv

stephenfuqua:
The problem noted by Luis above was an environmental issue in the Schoology setup, which I have now corrected.

lalrg (approved):
Code looks fine, tests passing. I run the tool and the output is the expected.

PR #76, opened 2020-11-05
[FIZZ-78] Logging improvements

stephenfuqua:
While intended primarily for Google Classroom, I made a few opportunistic updates in Schoology to keep them aligned.

### DIFF
--- a/src/google-classroom-extractor/google_classroom_extractor/csv_generation/write.py
+++ b/src/google-classroom-extractor/google_classroom_extractor/csv_generation/write.py
@@ -43,7 +43,6 @@ def write_csv(df_to_write: DataFrame, output_date: datetime, directory: str):
         logger.info(f"Generated file => {path}")
     except Exception:
         logger.exception("An exception occurred while writing file %s", path)
-        raise
 
 
 def write_multi_csv(
--- a/src/google-classroom-extractor/google_classroom_extractor/csv_generation/write.py
+++ b/src/google-classroom-extractor/google_classroom_extractor/csv_generation/write.py
@@ -3,6 +3,7 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
+import logging
 from typing import Dict, Tuple
 import os
 from datetime import datetime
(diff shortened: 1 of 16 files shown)


src/google-classroom-extractor/google_classroom_extractor/csv_generation/write.py
```diff
@@ -30,9 +33,16 @@ def write_csv(df_to_write: DataFrame, output_date: datetime, directory: str):
...
-    df_to_write.to_csv(os.path.join(directory, f"{filename}.csv"), index=False)
+    path = os.path.join(directory, f"{filename}.csv")
+
+    try:
+        os.makedirs(directory, exist_ok=True)
+        df_to_write.to_csv(path, index=False)
+
+        logger.info(f"Generated file => {path}")
+    except Exception:
+        logger.exception("An exception occurred while writing file %s", path)
```
stephenfuqua, line 45: This should really have a unit test - not covering the logging itself, but covering that the exception is caught here _and not bubbled up_. Why swallow the exception? Because if we bubble it up, then the entire application will likely stop running - whereas we want it to try running the next file in the list.

src/google-classroom-extractor/google_classroom_extractor/main.py
```diff
@@ -35,36 +42,51 @@
...
 
-def request() -> Result:
-    load_dotenv()
+
+def configure_logging():
+    global logger
+
+    logger = logging.getLogger(__name__)
+
+    level = os.environ.get("LOGLEVEL", "INFO")
```
stephenfuqua, line 53: I added a little more organization in this file. Schoology needs it too, but it would be more invasive than the other little refactors I've done, so I don't want to tackle that in this ticket.

src/google-classroom-extractor/google_classroom_extractor/main.py
```diff
@@ -35,36 +42,51 @@
...
         handlers=[
             logging.StreamHandler(sys.stdout),
         ],
-        format="%(asctime)s - %(levelname)s - %(message)s",
-        level=os.environ.get("LOGLEVEL", "INFO"),
+        format="%(asctime)s - %(levelname)s - %(name)s - %(message)s",
+        level=level,
     )
-    logging.getLogger("matplotlib").setLevel(logging.ERROR)
-    logging.getLogger("opnieuw").setLevel(logging.DEBUG)
```
stephenfuqua, line 49: @bradbanister do you agree that we don't need matplotlib here at all, and that opnieuw should log at the level that we just set in the `basicConfig(...)` call?

bradbanister: yeah, that's fine 👍

bradbanister (approved):
This all looks good.

PR #77, opened 2020-11-09
FIZZ-103

lalrg:


src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -87,6 +87,11 @@ def _get_users() -> pd.DataFrame:
 result_bucket: Dict[str, pd.DataFrame] = {}
 
 
+# There are some methods that will need to share data in order to follow the
+# pattern used by the `_create_file_from_dataframe` method.
+shared_data: Dict = {}
```
stephenfuqua, line 92: Why not just use the existing `result_bucket` variable? Can extend the type of the value instead of only typing it to `pd.DataFrame`. However, the name `shared_data` is better than `result_bucket` in that case.

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -115,9 +120,10 @@ def __get_user_activities() -> Optional[pd.DataFrame]:
     return __get_user_activities
 
 
-def _get_submissions() -> list:
-    assignments_df: pd.DataFrame = result_bucket["assignments"]
-    return service.get_submissions(assignments_df)
+def _get_submissions() -> pd.DataFrame:
+    assigment_id: int = shared_data["current_assignment_id"]
+    section_id: int = shared_data["current_section_id"]
```
stephenfuqua, line 125: The pattern used elsewhere is to pass these values as arguments instead of relying on cached values.

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -165,8 +171,14 @@ def main():
...
-            # TODO: use correct file path, and use DatFrame instead of list, in FIZZ-103
-            _create_file_from_list(_get_submissions, "submissions.csv")
+            assignments : pd.DataFrame = result_bucket["assignments"]
+
+            for assignment in assignments["SourceSystemIdentifier"].tolist():
+                shared_data["current_assignment_id"] = assignment
+                shared_data["current_section_id"] = section_id
+                assignment_id = int(assignment)
+                submission_file_name = lms.get_submissions_file_path(schoology_output_path, section_id, assignment_id)
+                _create_file_from_dataframe(_get_submissions, submission_file_name)
```
stephenfuqua, line 181: It looks like you should be able to use the existing pattern:

```suggestion
                _create_file_from_dataframe(_get_submissions(section_id, assignment_id), submission_file_name)
```

src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
```diff
@@ -154,7 +154,7 @@ def get_assignments(self, section_id: int) -> pd.DataFrame:
 
         return assignmentsMap.map_to_udm(assignments, section_id)
 
-    def get_submissions(self, assignments: pd.DataFrame) -> list:
+    def get_submissions(self, assignment_id: int, section_id: int) -> pd.DataFrame:
         """
         Gets all Schoology submissions for the given assignments.
```
stephenfuqua, line 159: Don't forget to update the docstring.

stephenfuqua (changes requested):

src/schoology-extractor/schoology_extractor/main.py
lalrg: I might need to change the typing. Dataframe does not work for my functionality... If you are okay with it, I can use a single `bucket`

src/schoology-extractor/schoology_extractor/main.py
lalrg: You are right... I'll change it. Thanks!

src/schoology-extractor/schoology_extractor/main.py
lalrg: I won't use the new shared data, just removed it

src/schoology-extractor/schoology_extractor/main.py
lalrg: I just changed the logic to make it work like the rest of methods

src/schoology-extractor/schoology_extractor/mapping/submissions.py
```diff
@@ -0,0 +1,80 @@
...
+
+    df["SourceSystem"] = constants.SOURCE_SYSTEM
+    df["EntityStatus"] = constants.ACTIVE
+    df["SubmissionStatus"] = df["late"].apply(lambda x: 'late' if x == 1 else '')
+    df["SubmissionStatus"] = df["draft"].apply(lambda x: 'draft' if x == 1 else x)
+    df.drop(columns=["late", "draft"], inplace=True)
+
+    df["AssignmentSourceSystemIdentifier"] = df["id"].apply(lambda x: x.split('#')[1])
+    df["EarnedPoints"] = None
+    df["Grade"] = None
```
stephenfuqua, line 67: Although I've looked at the Submissions API many times and I knew it didn't have grades, part of my brain kept thinking that we _would_ have grades when we got to this point. Makes me realize that we may have a gap in our upcoming stories, as the story on the Grades was about the _course grade_, not the _assignment grade_. I will look at the tickets and make sure we have a placeholder for filling this in.

src/schoology-extractor/schoology_extractor/mapping/submissions.py
```diff
@@ -0,0 +1,80 @@
...
+
+    df.rename(
+        columns={
+            "id": "SourceSystemIdentifier",
+            "created": "SubmissionDateTime",
```
stephenfuqua, line 72: Is `created` a Unix timestamp? If so, it needs to be converted to a "normal" DateTime string with format like `YYYY-mm-dd HH:MM:SS`

stephenfuqua: Never mind, I see it below on line 78 😁

src/schoology-extractor/schoology_extractor/mapping/submissions.py
```diff
@@ -0,0 +1,80 @@
...
+            "uid",
+            "CreateDate",
+            "LastModifiedDate"
+        ]
+    ].copy()
+
+    df["SourceSystem"] = constants.SOURCE_SYSTEM
+    df["EntityStatus"] = constants.ACTIVE
+    df["SubmissionStatus"] = df["late"].apply(lambda x: 'late' if x == 1 else '')
+    df["SubmissionStatus"] = df["draft"].apply(lambda x: 'draft' if x == 1 else x)
```
stephenfuqua, line 62: Is the `SubmissionStatus` going to be blank if it is not late or draft? Seems like it would be better to write "complete" or "on-time" here instead of leaving an ambiguous blank.

lalrg: ✅

src/schoology-extractor/schoology_extractor/mapping/submissions.py
```diff
@@ -58,8 +58,11 @@ def map_to_udm(submissions_df: pd.DataFrame) -> pd.DataFrame:
 
     df["SourceSystem"] = constants.SOURCE_SYSTEM
     df["EntityStatus"] = constants.ACTIVE
-    df["SubmissionStatus"] = df["late"].apply(lambda x: 'late' if x == 1 else '')
-    df["SubmissionStatus"] = df["draft"].apply(lambda x: 'draft' if x == 1 else x)
+    df["SubmissionStatus"] = 'on-time'
+    for index, row in df.iterrows():
+        row["SubmissionStatus"] = 'late' if row["late"] == 1 else row["SubmissionStatus"]
+        row["SubmissionStatus"] = 'draft' if row["draft"] == 1 else row["SubmissionStatus"]
```
stephenfuqua, line 64: Iterating rows should be avoided when possible.

```suggestion
       def _get_status(row: pd.Series):
           if row["late"] == 1:
               return "late"
           if row["draft"] == 1:
               return "draft"
           return "on-time"

        row["SubmissionStatus"] = df.apply(_get_status, axis=1)
```

Apply should be much faster. There are other possible optimizations but I haven't quite understood them yet.

lalrg: ✅

stephenfuqua (approved):

PR #78, opened 2020-11-10
[FIZZ-130] Add sync support to attendance_events

lalrg:


src/schoology-extractor/schoology_extractor/schoology_extract_facade.py
```diff
@@ -235,7 +235,24 @@ def get_attendance_events(
...
         events = self._client.get_attendance(section_id)
 
-        events_df = attendanceMap.map_to_udm(events, section_associations)
+        def _sync_wrapper(data: pd.DataFrame):
+            """ Attendance_events entity has a more complex mapping, for most
+            entities, sync process running before mapping is fine, but considering
+            the process involved for this entity, the mapper for attendance events
+            will accept an additional callable for sync process that will run when
+            the mapping process is ready
+            """
```
lalrg, line 244: I don't know if it is a good idea to have it as a docstring or I should just clarify it as a comment in the PR and remove it from the code. Please let me know your thoughts @stephenfuqua

stephenfuqua: Docstrings are used for Intellisense or for auto-generated documentation. If you think the comment should show up in those two situations, then I would use the docstring - but then also document the argument and return type. Otherwise I would just do a regular comment with `#` at the beginning of every line.

The signature needs the return type.

stephenfuqua: But this is probably fine.

src/schoology-extractor/README.md
```diff
@@ -3,6 +3,12 @@
 This script retrieves and writes out to CSV all students, sections, assignments,
 and submissions for the desired grading period(s).
 
+## Notes on Working With Schoology Data
+
+* The Schoology API handles _negative attendance_ events: if a student is marked
+  as present, or is not marked at all, then the system will not return a record
+  for that day.
```
stephenfuqua, line 10: I was surprised to realize this point and this felt like the best way to keep it in mind.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
```diff
@@ -40,7 +43,11 @@ def _get_status(status_code: int) -> str:
     return switcher.get(status_code, f"Unknown status: {status_code}")
 
 
-def map_to_udm(attendance: list, section_associations: pd.DataFrame) -> pd.DataFrame:
+def map_to_udm(
+    attendance: list,
+    section_associations: pd.DataFrame,
+    sync_callback: Callable = None,
```
stephenfuqua, line 49: Since the "additional mapping" is only for one purpose - not generic - I renamed the argument to `sync_callback`.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
```diff
@@ -93,7 +102,7 @@ def map_to_udm(attendance: list, section_associations: pd.DataFrame) -> pd.DataF
     # This data type conversion was required because Schoology is returning
     # enrollment Id as an integer in the Attendance endpoint, but as a string
     # with the Enrollment endpoint.
-    sa = sa.astype({'SourceSystemIdentifier': 'int64'})
+    sa = sa.astype({"SourceSystemIdentifier": "int64"})
```
stephenfuqua, line 105: I reformatted the file with `black`, by the way.

src/schoology-extractor/schoology_extractor/mapping/attendance.py
```diff
@@ -115,7 +124,7 @@ def map_to_udm(attendance: list, section_associations: pd.DataFrame) -> pd.DataF
         ]
     ]
 
-    df["CreateDate"] = None
-    df["LastModifiedDate"] = None
+    if sync_callback is not None:
+        df = sync_callback(df)
```
stephenfuqua, line 128: Something very important was missed before. Previously you were calling `sync_callback` but ignoring the result, so the output files did not have the `CreateDate` and `LastModifiedDate` columns.

stephenfuqua (approved):

PR #79, opened 2020-11-11
[FIZZ-131] Sync user activities

lalrg:


bradbanister (approved):
Code looks good, passed testing.

PR #83, opened 2020-11-16
[FIZZ-23] Import Usage Analytics

stephenfuqua:


src/schoology-extractor/schoology_extractor/usage_analytics_facade.py
```diff
@@ -0,0 +1,36 @@
...
+    #   output = pd.DataFrame()
+    #   for file in os.scandir({only .csv or .csv.gz files}):
+    #       raw_data = csv_reader.load_data_frame(file)
+    #       mapped_data = usageMap.map_to_udm(raw_data)
+    #       output.append(mapped_data)
+    #
+    #   # If reports have overlapping dates then we'll have duplicates
+    #   output.drop_duplicates(inplace=True)
+    #
+    #   return output
```
stephenfuqua, line 34: The above comments provide an outline of the remaining work for FIZZ-23.

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -40,21 +41,30 @@ def _get_current_date_with_format() -> str:
     return datetime.strftime(datetime.now(), "%Y-%m-%d %H:%M:%S")
 
 
+def _map_result_to_dicts_arr(result: Union[list[ResultProxy], None]):
```
stephenfuqua, line 44: Return type?

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -40,21 +41,30 @@ def _get_current_date_with_format() -> str:
...
+    mapped_arr: list[dict] = list[dict]()
+    if result is None:
+        return list()
```
stephenfuqua, line 47: Both uses of this function are followed by

```
db_item = dict() if len(mapped_result) == 0 else mapped_result[0]
```

Why not handle that directly in the function?

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -40,21 +41,30 @@ def _get_current_date_with_format() -> str:
...
+    for row in result:
+        db_item = dict()
+        for column, db_value in row.items():  # type: ignore
+            db_item[column] = db_value
+        mapped_arr.append(db_item)
```
stephenfuqua, line 52: You might try something like this instead:

```
mapped_arr = [dict(row.items()) for row in result]
```

stephenfuqua: Your code seems fine by the way. Just might be a small improvement, especially with larger results, to use list comprehension instead and take better advantage of SQL Alchemy.

src/schoology-extractor/schoology_extractor/helpers/sync.py
stephenfuqua: Oh, I see the point now - this function is returning all rows, but in the code you know that you only care about the first result and thus extract it with `[0]`. Makes sense now.

Might be useful to write one line method to encapsulate this, e.g. `_get_first(result_list)` or something like that.

src/schoology-extractor/schoology_extractor/helpers/sync.py
```diff
@@ -204,3 +214,63 @@ def sync_resource(
...
+
+    Returns
+    -------
+    bool
+        True if the file has been processed, False if it has not.
+    """
+
+    _create_usage_table_if_it_does_not_exist(db_engine)
+
+    query = f"SELECT * FROM {USAGE_TABLE_NAME} where FILE_NAME='{file_name}' limit 1;"
```
stephenfuqua, line 249: I just did a little experiment and found that the following syntax will work. This is a random query that I wrote just to give the idea. The nice thing is you always get a single row back, with value of 1 or 1. You could adapt this to simplify the code a little bit. But I would save this for tomorrow, after everything else already working well. Get the whole app working now and then optimize tomorrow.
```
select exists(select 1 from assignment where assignment.SourceId = 1)
```

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -144,6 +147,16 @@ def __get_attendance_events() -> pd.DataFrame:
     return __get_attendance_events
 
 
+def _get_system_activities(input_directory: str) -> Callable:
+    def __get_system_activities() -> pd.DataFrame:
+        system_activities = analyticsFacade.get_system_activities(input_directory)
+        result_bucket["system_activities"] = system_activities
```
stephenfuqua, line 153: I never see `result_bucket["system_activities"]` being used it again - is there a plan to access it, or perhaps can we remove that assignment altogether?

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -196,6 +209,13 @@ def main():
         )
         _create_file_from_dataframe(_get_attendance_events(section_id), file_path)
 
+    if input_directory is not None:
```
stephenfuqua, line 212: To improve readability, what do you think about adding an explanatory variable?

```suggestion
    need_to_process_input_files = input_directory is not None
    if need_to_process_input_files:
```

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -144,6 +147,16 @@ def __get_attendance_events() -> pd.DataFrame:
     return __get_attendance_events
 
 
+def _get_system_activities(input_directory: str) -> Callable:
```
stephenfuqua, line 150: `input_directory` is a class variable, so there's really no need to pass it to the function. And because you don't need to pass it to the function, you don't need to create closure over it - so you could eliminate the inner function.

src/schoology-extractor/schoology_extractor/main.py
```diff
@@ -196,6 +209,13 @@ def main():
         )
         _create_file_from_dataframe(_get_attendance_events(section_id), file_path)
 
+    if input_directory is not None:
+        system_activities_output_dir = lms.get_system_activities_file_path(schoology_output_path)
+        _create_file_from_dataframe(
+            _get_system_activities(input_directory),
```
stephenfuqua, line 215: Once the inner function is eliminated, just pass the function name here without actually calling the function.
```suggestion
            _get_system_activities,
```

src/schoology-extractor/schoology_extractor/usage_analytics_facade.py
```diff
@@ -3,34 +3,75 @@
...
+        Database connectivity for sync process
+    """
+    db_engine: sqlalchemy.engine.base.Engine
+
+    @property
+    def _db_engine(self):
+        assert isinstance(self.db_engine, sqlalchemy.engine.base.Engine)
+        return self.db_engine
+
+    def get_system_activities(self, usage_input_dir: str) -> pd.DataFrame:
```
stephenfuqua, line 39: I was thinking about this module being purely functional, rather than being a class. Since there is only one input argument to the class (the `db_engine`) and there is really no need to preserve state, it would be simple to just pass that argument in as an argument `get_system_activities`.

src/schoology-extractor/schoology_extractor/usage_analytics_facade.py
```diff
@@ -3,34 +3,75 @@
...
+        logger.debug(f"Processing usage_analytics files: loading files from {usage_input_dir}")
+        for file in os.scandir(usage_input_dir):
+            # It is not expected to have anything different from .gz or .csv
+            # in case there's something different, this method will throw an
+            # exception
+            logger.debug(f"Processing usage_analytics files: processing {file.name}")
+            row_data = csv_reader.load_data_frame(file.path)
+            mapped_row = pd.DataFrame()
+
+            if not sync.usage_file_is_processed(file.name, self._db_engine):
```
stephenfuqua, line 64: Shouldn't the check to see if the file has been processed be run before reading the file? it might be nice to avoid the overhead of parsing the input file if we don't need the results.

src/schoology-extractor/schoology_extractor/main.py
lalrg: I can remove it.

src/schoology-extractor/schoology_extractor/usage_analytics_facade.py
lalrg: It is easy to make it functional. I just wrote the function to have similar apis for the two facades. I can change it if needed.

stephenfuqua: Let's prefer functional at this point, now that we're getting used to that paradigm.

src/schoology-extractor/schoology_extractor/helpers/sync.py
lalrg: ✔️

src/schoology-extractor/schoology_extractor/helpers/sync.py
lalrg: ✔️

src/schoology-extractor/schoology_extractor/main.py
lalrg: ✔️

src/schoology-extractor/schoology_extractor/helpers/sync.py
lalrg: ✔️

src/schoology-extractor/schoology_extractor/main.py
lalrg: ✔️

src/schoology-extractor/schoology_extractor/main.py
lalrg: ✔️

src/schoology-extractor/schoology_extractor/main.py
lalrg: ✔️

src/schoology-extractor/schoology_extractor/usage_analytics_facade.py
lalrg: ✔️

src/schoology-extractor/schoology_extractor/usage_analytics_facade.py
lalrg: ✔️

stephenfuqua:
Closing this without merge so that @lalrg can open a new one.

PR #86, opened 2020-11-18
[FIZZ-23] Export Schoology System Sign-In and Sign-Out

lalrg:


stephenfuqua:
Code and manual testing look good. Holding off on approval and merge because the coverage utility is failing when running one of the unit tests, presumably because that unit test mocked out an OS level function.

stephenfuqua (approved):

PR #87, opened 2020-11-19
[FIZZ-91] factor out sync-specific code from courses

bradbanister:
A PR for the factoring out of sync/CreateDate/LastModifiedDate, the first step of FIZZ-91, to allow for others to pick up GC stories if they want.

### DIFF
--- a/src/google-classroom-extractor/google_classroom_extractor/api/courses.py
+++ b/src/google-classroom-extractor/google_classroom_extractor/api/courses.py
@@ -4,13 +4,15 @@
 # See the LICENSE and NOTICES files in the project root for more information.
 
 import logging
-from datetime import datetime
 from typing import List, Dict, Optional, cast
-from pandas import DataFrame, json_normalize, read_sql_query
+from pandas import DataFrame, json_normalize
 import sqlalchemy
-import xxhash
 from googleapiclient.discovery import Resource
 from google_classroom_extractor.api.api_caller import call_api, ResourceType
+from google_classroom_extractor.api.resource_sync import (
+    cleanup_after_sync,
+    sync_to_db_without_cleanup,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -124,15 +126,14 @@ def request_all_courses_as_df(
 
     courses_df = request_latest_courses_as_df(resource)
     _sync_courses_without_cleanup(courses_df, sync_db)
+    cleanup_after_sync("Courses", sync_db)
 
-    # finally, remove temporary sync tables
-    with sync_db.connect() as con:
-        con.execute("DROP TABLE IF EXISTS Sync_Courses")
-        con.execute("DROP TABLE IF EXISTS Unmatched")
     return courses_df
 
 
-def _sync_courses_without_cleanup(courses_df: DataFrame, sync_db: sqlalchemy.engine.base.Engine):
+def _sync_courses_without_cleanup(
+    courses_df: DataFrame, sync_db: sqlalchemy.engine.base.Engine
+):
     """
     Take fetched API data and sync with database. Creates tables when necessary,
     but ok if temporary tables are there to start. Doesn't delete temporary tables when finished.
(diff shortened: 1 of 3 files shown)


src/google-classroom-extractor/google_classroom_extractor/api/resource_sync.py
```diff
@@ -0,0 +1,241 @@
+# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: This is the extracted code

src/google-classroom-extractor/google_classroom_extractor/api/resource_sync.py
```diff
@@ -0,0 +1,241 @@
...
+        the name of the API resource, e.g. "Courses", to be used in SQL
+    sync_db: sqlalchemy.engine.base.Engine
+        an Engine instance for creating database connections
+    """
+    with sync_db.connect() as con:
+        con.execute(f"DROP TABLE IF EXISTS Sync_{resource_name}")
+        con.execute(f"DROP TABLE IF EXISTS Unmatched_{resource_name}")
+
+
+def sync_to_db_without_cleanup(
```
bradbanister, line 31: I split out db temp table cleanup for testability. Integration tests can confirm that the intermediate work done in the temp tables is as expected.

src/google-classroom-extractor/google_classroom_extractor/api/resource_sync.py
```diff
@@ -0,0 +1,241 @@
...
+        # set up index for update operation
+        for pk in primary_keys_as_array:
+            create_date_df[pk] = create_date_df[pk].astype("string")
+        create_date_df.set_index(primary_keys_as_array, inplace=True)
+
+        # update CreateDate/LastModifiedDate of sync records
+        resource_df.update(create_date_df)
+
+        # reset index so no columns are hidden
+        resource_df.reset_index(inplace=True)
```
stephenfuqua, line 241: This function is very long - much longer than I usually like to see. Generally I've followed OO-guru refactoring advice: instead of a comment describing a few lines of code, extract those lines into a small, private (possibly local), well-named function. Makes it easy to read the public function in a few lines - basically like skimming for just the comments.

On the other hand, a lot of the lines in this case are for SQL statements and formatting. And honestly the cognitive load to read this function is not particularly high, since each individual step is easy to understand.

I'm commenting on this, but in the mode of discussion rather than requesting a change.

src/google-classroom-extractor/google_classroom_extractor/api/courses.py
```diff
@@ -4,13 +4,15 @@
...
 import logging
-from datetime import datetime
 from typing import List, Dict, Optional, cast
-from pandas import DataFrame, json_normalize, read_sql_query
+from pandas import DataFrame, json_normalize
 import sqlalchemy
-import xxhash
 from googleapiclient.discovery import Resource
 from google_classroom_extractor.api.api_caller import call_api, ResourceType
+from google_classroom_extractor.api.resource_sync import (
```
lalrg, line 12:
```suggestion
from googleapiclient.discovery import Resource

from google_classroom_extractor.api.api_caller import call_api, ResourceType
from google_classroom_extractor.api.resource_sync import (
```

src/google-classroom-extractor/google_classroom_extractor/api/resource_sync.py
```diff
@@ -0,0 +1,241 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import logging
+from datetime import datetime
+from pandas import DataFrame, read_sql_query
+import sqlalchemy
+import xxhash
```
lalrg, line 10:
```suggestion
import logging
from datetime import datetime

from pandas import DataFrame, read_sql_query
import sqlalchemy
import xxhash
```

src/google-classroom-extractor/google_classroom_extractor/api/resource_sync.py
lalrg: I'd also create small local functions, even when it is not very hard to read, would be easier if we had at the end something like:
```python3
def sync_to_db_without_cleanup():
    # functions definition
    # ...
    # ...
    create_sync_table()
    prepare_df()
    set_created_and_modified_date()
    # other methods calls
```

lalrg: I am not sure if this should be done now(or later)... This is just what I think, answering Stephen's comment

stephenfuqua: Good example of what the function would end up looking like, if refactored. Except each of those would want underscores at the beginning of the function name 😁 .

bradbanister: yep, i can do that

stephenfuqua (approved):
Approving from my perspective, but leaving open for @lalrg's review.

lalrg (approved):
LGTM

PR #90, opened 2020-11-19
[FIZZ-91] Implement Student and Teacher sync with Users output of CreateDate/LastModifiedDate

bradbanister:


lalrg (changes requested):
Just a small change. Code loks good in general

lalrg (approved):
Code looks good.
I tested the functionality for users sync and it works fine.

PR #91, opened 2020-11-20
[FIZZ-156] CreateDate/LastModifiedDate on user section associations

bradbanister:


stephenfuqua (approved):
Nice! so little work to do here 😄 . Leaving this open as @lalrg  appears to be doing QA for FIZZ-156.

lalrg (approved):
Code looks good. I tested the functionality and it works as expected.

PR #92, opened 2020-11-20
[FIZZ-154] Assignments sync

lalrg:


src/google-classroom-extractor/google_classroom_extractor/api/coursework.py
```diff
@@ -170,18 +176,56 @@ def request_all_coursework_as_df(
...
+            "dueDate.day" TEXT,
+            "dueTime.hours" TEXT,
+            "dueTime.minutes" TEXT,
+            "assignment.studentWorkFolder.id" TEXT,
+            "assignment.studentWorkFolder.title" TEXT,
+            "assignment.studentWorkFolder.alternateLink" TEXT,
+            topicId TEXT,
+            scheduledTime TEXT,
+            """,
+        primary_keys="id",
```
bradbanister, line 229: So, "id" probably isn't enough for identity for any of the other tickets.  In this case, it's courseId and id together because id is only unique per course.  See https://developers.google.com/classroom/reference/rest/v1/courses.courseWork
You can cheat by looking at the mapping code where I join id fields with "-" to see which ones are needed for uniqueness.

lalrg: Thank you

stephenfuqua (approved):

PR #96, opened 2020-12-01
[FIZZ-159] Small UX improvements

lalrg:
Changes approved by Eric


ejansson:
Tested this and the UX works as expected.

src/schoology-extractor/README.md
```diff
@@ -40,6 +40,8 @@ to `.env`, and adjust to your desired parameters. Supported parameters:
 User's API key and secret are managed at
 [https://app.schoology.com/api](https://app.schoology.com/api).
 
+For more information about log levels [click here](https://techdocs.ed-fi.org/display/EDFIODS/Fizz+-+Exception+Handling+and+Logging)
```
stephenfuqua, line 43: 1. This page is only accessible to Ed-Fi developers and should not be referenced here. Let's just list the acceptable log levels here in a table without reproducing all of the details from that page.
1. General comment: I prefer to avoid "click here" type links. Better to have something like `See [Fizz Exception Handling and Logging](..) for more information on log levels`. This provides more useful information than the generic text "click here".

stephenfuqua (changes requested):

src/schoology-extractor/README.md
lalrg: ✔️

stephenfuqua (approved):

PR #98, opened 2020-12-02
[FIZZ-150] Basic error handler

lalrg:


src/google-classroom-extractor/google_classroom_extractor/main.py
```diff
@@ -134,6 +138,11 @@ def main():
 
     logger.info("Finishing Ed-Fi LMS Google Classroom Extractor")
 
+    if error_tracker.fired:
+        print("fatal error", file=sys.stderr)
```
stephenfuqua, line 142:
```suggestion
        print("A fatal error occurred, please review the log output for more information.", file=sys.stderr)
```

stephenfuqua (changes requested):
I've generated two different errors and it is not triggering this process. Investigating still.

stephenfuqua:
> I've generated two different errors and it is not triggering this process. Investigating still.

For example, set this environment variable before running, pointing to a certificate that doesn't exist on your computer: `$env:HTTPLIB2_CA_CERTS="c:\msdfrootca.cert"`

I also have an error because of something in SQLite, which will probably resolve if I delete my local database and start over.

src/google-classroom-extractor/google_classroom_extractor/main.py
```diff
@@ -134,6 +138,11 @@ def main():
...
+        sys.exit(-1)
```
stephenfuqua, line 143:
```suggestion
        sys.exit(1)
```

stephenfuqua (approved):

PR #99, opened 2020-12-02
[FIZZ-187] Bulk create Schoology users in API

bradbanister:
Making this a draft PR because I need to add a README.  To run:
`poetry run python tests/data_generation/main.py`
Edit main.py `NUMBER_OF_USERS` to change number to generate.
As-is, the code rolls back users after creating them.  See users.py line 123


src/schoology-extractor/tests/data_generation/main.py
```diff
@@ -0,0 +1,28 @@
+# SPDX-License-Identifier: Apache-2.0
```
stephenfuqua, line 1: Is the plan to move this to `/utils/something`? Alternately could leave it in the extractor as `test_data_generatory.py` sitting next to `main.py`, if we think that downstream users might be interested and if that simplifies package / module references.

src/schoology-extractor/tests/data_generation/users.py
```diff
@@ -0,0 +1,145 @@
...
+from http import HTTPStatus
+from faker import Faker
+from schoology_extractor.api.request_client import RequestClient
+
+fake = Faker("en_US")
+logger = logging.getLogger(__name__)
+
+VALID_SCHOOL_ID = "2908525646"
+VALID_ROLE_ID = "796380"
+VALID_EMAIL_DOMAIN = "studentgps.org"
```
stephenfuqua, line 17: This makes sense. Also calls into question my comment above about making available to downstream users right in `schoology_extractor`. Separate directory is definitely better. I can live with the current directory structure, even though it  is a bit "odd".

src/schoology-extractor/tests/data_generation/users.py
```diff
@@ -0,0 +1,145 @@
...
+        A RequestClient initialized for access to the Schoology API
+    record_count : int
+        The number of users to generate.
+    """
+    assert record_count > 0, "Number of users to generate must be greater than zero"
+
+    total_users: List[Dict] = generate_users(record_count)
+
+    MAX_CHUNK_SIZE = 50
+    chunked_users: List[List[Dict]] = [total_users[x:x+MAX_CHUNK_SIZE] for x in range(0, len(total_users), MAX_CHUNK_SIZE)]
```
stephenfuqua, line 142: I still get a kick out of writing / seeing list comprehension and slicing in Python 😆.

src/schoology-extractor/tests/data_generation/main.py
bradbanister: My thinking behind having it under `tests` is that it's essentially a generator of test data.  Not the best fit, but a reasonable first cut at a location.

src/schoology-extractor/tests/data_generation/users.py
bradbanister: I know!  My initial thought for chunking was to write a reducer, but I'm sure that's too functional-oriented to be "pythonic". 🙃

src/schoology-extractor/tests/data_generation/main.py
```diff
@@ -0,0 +1,28 @@
...
+        logging.StreamHandler(stdout),
+    ],
+    format="%(asctime)s - %(levelname)s - %(name)s - %(message)s",
+    level="INFO",
+)
+logger = logging.getLogger(__name__)
+
+NUMBER_OF_USERS = 60
+
+generate_and_load_users(RequestClient(arguments.client_key, arguments.client_secret), NUMBER_OF_USERS)
```
stephenfuqua, line 28: As discussed in Slack, I'm going to change the file name to `__main__.py`.

stephenfuqua (approved):
Looks & works great.

PR #102, opened 2020-12-08
FIZZ-180 added lms function to pull system activity data and created …

ggaronzik:
…notebook to visualize those results


src/notebooks/lms_utilities.py
```diff
@@ -1,158 +1,170 @@
...
+def get_attendance_events(base_directory: str, section_id: int) -> pd.DataFrame:
+    file = get_attendance_events_file(base_directory, section_id)
+
+    if file is not None:
+        return _read_csv(file)
+
+    return pd.DataFrame()
+
+def get_all_attendance_events(base_directory: str, sections: pd.DataFrame) -> pd.DataFrame:
+    return _get_data_for_section(base_directory, sections, get_attendance_events)
```
stephenfuqua, line 170: This file no longer exists; you need to merge the `main` branch into your branch.

Please see the latest [filesystem tutorial](https://github.com/Ed-Fi-Exchange-OSS/Ed-Fi-X-Fizz/blob/main/src/notebooks/filesystem-tutorial.ipynb), which has code like this:

```python
# Load this helper module to access the functions
# Instead of loading from a published package, we're going to hack the 
# module search path to find the package on the local file system.
import sys
sys.path.append(os.path.join("..","file-utils"))

from lms_file_utils import file_repository # contains functions that get directory paths
from lms_file_utils import file_reader # contains the functions that load data frames
```

And now a function like `get_all_attendance_events` needs to be called with `file_reader.get_all_attendance_events`.

src/notebooks/student_logins.ipynb
```diff
@@ -0,0 +1,277 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# add user and activity columns to login frequency dataframe\n",
+    "login_df = pd.DataFrame()\n",
+    "login_df = user_logins_df.loc[:,['Name', 'EmailAddress', 'SISUserIdentifier','ActivityType']].query(\"ActivityType not in ['sign-out']\")\n",
```
stephenfuqua, line 57: You can simplify a little: since you want all rows, it is unnecessary to use `.loc`:
```suggestion
    "login_df = user_logins_df.[['Name', 'EmailAddress', 'SISUserIdentifier','ActivityType']].query(\"ActivityType not in ['sign-out']\")\n",
```

src/notebooks/student_logins.ipynb
```diff
@@ -0,0 +1,277 @@
...
+    "\n",
+    "# drop ActivityType column\n",
+    "del login_df['ActivityType']\n",
```
stephenfuqua, line 60: This works just fine. I _think_ - but not 100% sure - that it would be more efficient to use the `drop` function though:

```suggestion
    "login_df.drop(columns=['ActivityType'], inplace=True)\n",
```

stephenfuqua: Interestingly, my assumption is not correct. For the code below, I get the following timings:

* `del` -> 737 microseconds
* `drop` -> 1.01 milliseconds
* `drop(inplace=True)` -> 1 milleseconds

No significant difference on the two `drops` but over 20% faster to use `del`. At that speed it hardly matters in this case :-D. So `del` is great. I'll probably stick with `drop` as is it idiomatically appropriate for a DataFrame.

```python
import pandas as pd
a=list()
b=list()
for i in range(100000):
    a.append(i)
    b.append(-i)

c = zip(a,b)
df = pd.DataFrame(c, columns=["a", "b"])

def _del():
    df1 = df.copy()
    del df1["b"]

def _drop():
    df2 = df.copy()
    df2 = df2.drop(columns=["b"])

def _drop_inplace():
    df3 = df.copy()
    df3.drop(columns=["b"], inplace=True)

%timeit _del()
%timeit _drop()
%timeit _drop_inplace()
```

src/notebooks/student_logins.ipynb
```diff
@@ -0,0 +1,277 @@
+{
+ "cells": [
```
stephenfuqua, line 2: Can you add a markdown cell at the top describing the purpose of this notebook, and the data that it will analyze?

stephenfuqua (changes requested):

src/notebooks/student_logins.ipynb
ggaronzik: i originally did not include the .loc but got a warning message recommending it.

.teamcity/_Self/FileTester/Project.kt
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
```
stephenfuqua: Why does this file show up? I'm disturbed by that...

stephenfuqua: It disappeared now. Weird.

stephenfuqua (approved):

PR #106, opened 2020-12-10
[FIZZ-194] Added enrollments, assignments, discussions and comments for Schoology generator

bradbanister:


src/schoology-extractor/tests/data_generation/__main__.py
```diff
@@ -90,6 +102,61 @@
...
+        enrollments=enrollments,
+    )
+except Exception as ex:
+    logger.exception(ex)
+
+
+try:
+    rollback_loaded_discussion_comments(request_client, discussion_comments)
+except Exception as ex:
+    logger.exception(ex)
```
stephenfuqua, line 145: When we're ready to do a real load of a large set of data, do you plan on just commenting out the rollbacks?

stephenfuqua (changes requested):
There's a TeamCity failure.

src/schoology-extractor/tests/data_generation/__main__.py
bradbanister: Yes.  It's in the readme, but I'll edit to better call it out.

src/schoology-extractor/schoology_extractor/api/request_client.py
```diff
@@ -167,13 +182,22 @@ def post(self, resource: str, json: dict) -> dict:
...
         )
 
+        if response.status_code == HTTPStatus.TOO_MANY_REQUESTS:
+            logger.warn(
+                "Hit API rate limiting on POST to %s, will retry with backoff", url
+            )
+            raise HTTPError(
+                f"{response.reason} ({response.status_code}): {response.text}"
+            )
+
```
stephenfuqua, line 200: Would it make sense to extract a function for this block of code, to avoid duplication?

stephenfuqua (changes requested):

stephenfuqua (approved):
Looks good now, thanks. And runs without errors. I saw a number of rate limit warnings while running, as expected, but it did run to conclusion without error.

PR #107, opened 2020-12-10
[FIZZ-185] Write empty files

stephenfuqua:


bradbanister (approved):
Looks good.  Runs as expected with temp data from new loader.

PR #109, opened 2020-12-10
[FIZZ-185] Ignore empty / "empty" csv files

stephenfuqua:
The tester now throws an error when it encounters "empty" files generated after FIZZ-185. These commits fix that, so that empty files are ignored when looking for the "newest" file in the filesystem.

Note that "empty" in this case actually means there are two line breaks, as `pandas.DataFrame.to_csv` writes empty DataFrames out that way.


bradbanister (approved):
Looks good.  Confirmed before and after load testing as expected.

PR #113, opened 2020-12-14
[FIZZ-195] Modify Google Classroom sync process to be generic

bradbanister:
Google Classroom sync process is now a function that takes a DataFrame, list of identity columns on the DataFrame, resource name (only for sync table naming) and sync DB engine.  All sync db tables have six columns:

SourceId - A composite identity for the data
Json - The data serialized to JSON
Hash - A hash of the data
CreateDate - The date the data was first created in the sync db
LastModifiedDate - The date the data was last updated in the sync db
SyncNeeded - A flag used during sync processing

The function returns a new DataFrame with the columns of the given DataFrame plus CreateDate and LastModifiedDate columns from the sync process.

### DIFF
--- a/src/google-classroom-extractor/.vscode/launch.json
+++ b/src/google-classroom-extractor/.vscode/launch.json
@@ -11,6 +11,12 @@
       "program": "${file}",
       "console": "integratedTerminal",
       "justMyCode": false
+    },
+    {
+      "name": "Python: Test debug config",
+      "type": "python",
+      "request": "test",
+      "console": "integratedTerminal"
     }
   ]
 }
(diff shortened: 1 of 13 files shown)


stephenfuqua (commented):
I like what I see.

In order to work with Schoology, we'll need to move `sync_to_db_without_cleanup` to a shared package. We can make that a task for a follow-up ticket for refactoring Schoology, or incorporate into a Canvas ticket. @bradbanister will you document that in Jira, in a new or existing ticket?

I'm leaving the final "Approve" action to @lalrg after he does testing.

lalrg (approved):
I did some testing and the code is working fine.

PR #115, opened 2020-12-15
[FIZZ-143] Update NOTICES.md

lalrg:


bradbanister (commented):
This looks good.  My only question is whether we need to have the dev-dependency libraries (e.g. pylint, pook, freezegun) in here.

stephenfuqua:
> This looks good. My only question is whether we need to have the dev-dependency libraries (e.g. pylint, pook, freezegun) in here.

Gray area. No harm in leaving them there.

stephenfuqua (approved):
I'm disgusted by the length of the GPL license text, which states that you must copy the whole thing. That implies to me that we must keep the lengthy preamble and the lengthy "how to use this license" in addition to the terms and conditions. We could have included just one copy here and linked to it from all projects that needed it, but that would have been more work for you.

I was also thinking about cutting down the Python portion because there are several licenses there, only half of which apply to us. But its just not worth it. We've created the file and met our legal obligations. No one actually cares to read these things anyway.

PR #116, opened 2020-12-15
[FIZZ-182] updated MetaEd files and UML diagram to current model

ggaronzik:


bradbanister (commented):
This looks good to me, but I'll hold off on immediate approval in case anyone else wants to chime in.

bradbanister (approved):

PR #118, opened 2020-12-17
[FIZZ-54] mirroring GC extractor structure, using canvasapi lib, using GC sync

bradbanister:
Draft PR with functionality complete.  Still working on tests.


src/canvas-extractor/canvas_extractor/api/canvas_helper.py
```diff
@@ -0,0 +1,86 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
```
lalrg, line 5: I prefer having a "helpers" directory for functions that are not calling the API.

src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -0,0 +1,49 @@
...
+from api.enrollments import enrollments_synced_as_df, request_enrollments
+from api.students import request_students, students_synced_as_df
+from api.submissions import request_submissions, submissions_synced_as_df
+from config import get_canvas_api, get_sync_db_engine
+# import export_data
+
+load_dotenv()
+
+canvas: Canvas = get_canvas_api()
+
```
lalrg, line 29: I prefer having the code inside functions, then calling them in `main()`. It could be separated by task(request, sync, export) or by resource(get_courses, get_students...)... In the case of separating by resource it could make all the calls for get, sync and export the resource.

src/canvas-extractor/canvas_extractor/api/canvas_helper.py
bradbanister: The to_df() should only be called by code in this directory.  The remove_duplicates() is more generic, but right now only code in this directory uses it.  We could split it off, but I'd wait to move it up a level until something outside this directory wants to use it.

src/canvas-extractor/canvas_extractor/__main__.py
bradbanister: Yes, it's pretty simplistic, I'll update.  Best would be to align on resource.  Hadn't organized code here at the start of the work because I didn't know what the dependencies would look like.  Turns out it's simple if you maintain lists of the canvasapi result objects.

bradbanister: i'll do this in my next PR

lalrg (approved):

PR #123, opened 2020-12-21
[FIZZ-74] Basic users mapping

lalrg:


src/canvas-extractor/canvas_extractor/mapping/users.py
```diff
@@ -0,0 +1,75 @@
...
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import pandas as pd
+
+from . import constants
+
+
+def map_to_udm(users_df: pd.DataFrame) -> pd.DataFrame:
```
bradbanister, line 11: Not a strong opinion, but you could rename to something like "map_to_udm_users" and then not have to rename when importing multiple of them.

bradbanister (commented):
Looking good so far

src/canvas-extractor/canvas_extractor/mapping/users.py
lalrg: ✔️

src/canvas-extractor/canvas_extractor/csv_generation/write.py
```diff
@@ -0,0 +1,91 @@
+# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: If this is identical to the GC version, maybe put a big comment to that effect so we know to extract to a shared library.  See api/resource_sync.py for an example.

bradbanister (approved):
Looks good, consider adding code comment if it applies.

src/canvas-extractor/canvas_extractor/csv_generation/write.py
lalrg: ✔️

PR #126, opened 2020-12-29
[FIZZ-133] Section-associations

lalrg:


src/canvas-extractor/canvas_extractor/mapping/section_associations.py
```diff
@@ -0,0 +1,123 @@
...
+# See the LICENSE and NOTICES files in the project root for more information.
+
+from datetime import datetime
+from pandas import DataFrame
+
+from .constants import SOURCE_SYSTEM, ACTIVE
+
+
+def _get_enrollment_status(status: str) -> str:
+    if status == "active":
```
bradbanister, line 13: Curious why these are being mapped.  Did we normalize enrollment status?

bradbanister (approved):
Looks good and runs as expected.  Minor question on mapping, but doesn't block approval.

src/canvas-extractor/canvas_extractor/mapping/section_associations.py
lalrg: I am just copying the statusses present in Schoology, but I did not find a standard for it

PR #127, opened 2020-12-31
[FIZZ-135] Map assignment-submissions

lalrg:


src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -127,14 +130,27 @@ def extract_assignments(
...
+            for assignment in assignments
+            if assignment.course_id == section.course_id
+        ]:
+            submissions: List[Submission] = request_submissions(assignment)
+            submissions_df: DataFrame = submissions_synced_as_df(submissions, sync_db)
+            submissions_df = map_to_udm_submissions(submissions_df)
+            export[(section.id, assignment.id)] = submissions_df
+
+    SUBMISSIONS_OUTPUT_DIRECTORY = os.path.join(
+        BASE_OUTPUT_DIRECTORY, SUBMISSION_ROOT_DIRECTORY
```
stephenfuqua, line 151: I would have expected the base output directory to come from configuration, rather than being a hard-coded constant.

Would it be possible to more closely follow the pattern of either Google Classroom or Schoology in determining the file system? In fact, the LmsFilesystemProvider from Schoology could be extracted into a shared module so that all extractors can use the same code.

src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -40,12 +40,15 @@
...
 from canvas_extractor.mapping.sections import map_to_udm_sections
+from canvas_extractor.mapping.submissions import map_to_udm_submissions
 from canvas_extractor.csv_generation.write import (
+    SUBMISSION_ROOT_DIRECTORY,
     write_csv,
     write_multi_csv,
     SECTIONS_ROOT_DIRECTORY,
     USERS_ROOT_DIRECTORY,
     ASSIGNMENT_ROOT_DIRECTORY,
+    write_multi_tuple_csv,
```
stephenfuqua, line 51: Is there any way to further differentiate the names `write_csv`, `write_multi_csv`, and `write_multi_tuple_csv` so that, when used below, it will be more obvious to the reader why one function was selected over the others?

src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -127,14 +130,27 @@ def extract_assignments(
...
+            submissions_df = map_to_udm_submissions(submissions_df)
+            export[(section.id, assignment.id)] = submissions_df
+
+    SUBMISSIONS_OUTPUT_DIRECTORY = os.path.join(
+        BASE_OUTPUT_DIRECTORY, SUBMISSION_ROOT_DIRECTORY
+    )
+    write_multi_tuple_csv(export, datetime.now(), SUBMISSIONS_OUTPUT_DIRECTORY)
 
 
 def extract_enrollments(
```
stephenfuqua, line 156: Would it make sense to move any of this code to a facade class?

src/canvas-extractor/canvas_extractor/__main__.py
bradbanister: A good idea.  Luis and I talked about using the Google Classroom code for writing data out, so that's where the naming comes from.  When we extract this out, we should improve the naming.

src/canvas-extractor/canvas_extractor/__main__.py
bradbanister: Now that these have multiplied in the main file, I'd suggest moving these functions to their own modules in a similar pattern to the "api" and "mapping" module groupings.  Maybe under an "extract" grouping?

bradbanister (changes requested):
overall looks good.  waiting on approval for discussion regarding minor restructure/rename.

src/canvas-extractor/canvas_extractor/__main__.py
lalrg: I like the idea of a facade

src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -127,14 +130,27 @@ def extract_assignments(
...
+            assignment
+            for assignment in assignments
+            if assignment.course_id == section.course_id
+        ]:
+            submissions: List[Submission] = request_submissions(assignment)
+            submissions_df: DataFrame = submissions_synced_as_df(submissions, sync_db)
+            submissions_df = map_to_udm_submissions(submissions_df)
+            export[(section.id, assignment.id)] = submissions_df
+
+    SUBMISSIONS_OUTPUT_DIRECTORY = os.path.join(
```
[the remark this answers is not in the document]
lalrg: ✔️

src/canvas-extractor/canvas_extractor/__main__.py
lalrg: it is currently using GC approach, I have just added the .env variable support... The reason why we choose the GC over Schoology is because of GC is more generic and easier to maintain since it does not add a new method each time that we need to introduce a new path, but we can discuss about it.
@bradbanister, what do you think about this point?

bradbanister: I think what's missing is the parse-args stuff in the other extractors.  I think it's in a separate story?

lalrg: That's part of FIZZ-56... the difference between the extractors regarding the filesystem, is that GC has a generic method and Schoology has a dedicated method for each file

src/canvas-extractor/canvas_extractor/__main__.py
lalrg: Is it okay to leave it for when we extract that logic in a shared library? or it should be done in this story?

bradbanister: I think it's ok to wait.  We should do the extract in the upcoming sprint.

src/canvas-extractor/canvas_extractor/__main__.py
stephenfuqua: Missing parse-arg stuff makes sense now. I'm not in a good spot right now for judging between the two (GC/Schoology) approaches, although my reaction-without-looking-at-code is that dedicated is better than generic, as it provides transparency to end-users and future programmers.

src/canvas-extractor/canvas_extractor/__main__.py
stephenfuqua: Please make sure that you write that as an acceptance criteria in an upcoming (or new) ticket.

PR #129, opened 2021-01-07
[FIZZ-75] Improve logging and exception handling

lalrg:


src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -70,65 +88,58 @@ def parse_args():
...
+    results_store["courses"] = (courses, courses_df)
 
 
-def _get_sections(
-    courses: List[Course], sync_db: sqlalchemy.engine.base.Engine
-) -> Tuple[List[Section], DataFrame]:
+@catch_exceptions
+def _get_sections(sync_db: sqlalchemy.engine.base.Engine) -> None:
     logger.info("Extracting Sections from Canvas API")
+    (courses, _) = results_store["courses"]
```
bradbanister, line 125: What's the thinking behind moving to using "results_store"?  I like being able to see all of a function's dependencies its parameter list.

lalrg: In the main function there were some assignments based on the return value from each function, if we stop returning them, we can controll what to do when a specific function fails, you can see the "succeeded" validations in the main method and how the decorator wraps the function and returns true or false depending on the resolution of the Callable.

src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -52,7 +48,29 @@
 start_date: str = ""
 end_date: str = ""
 
+results_store: Dict[str, Tuple[list, DataFrame]] = {}
```
bradbanister, line 51: The type of results_store needs to be loosened because on line 151, udm_assignments_df is a Dict[str, DataFrame], not a DataFrame:
`results_store: Dict[str, Tuple[list, Union[DataFrame, Dict[str, DataFrame]]]] = {}`

bradbanister: BTW, looking forward to Python 3.10 release, which includes TypeScript-like Union operator:
https://www.python.org/dev/peps/pep-0604/
Turns that into a more reasonable:
`results_store: Dict[str, Tuple[list, DataFrame | Dict[str, DataFrame]]] = {}`

lalrg: ✔️

bradbanister (approved):
Looks good!

PR #133, opened 2021-01-20
[FIZZ-199] new shared library package used by GC extractor

bradbanister:


bradbanister:
This is actually my FIZZ-200 spike work, but is ready to be carried forward for FIZZ-199.  See https://tracker.ed-fi.org/browse/FIZZ-200

stephenfuqua:
Looks nice so far. Do not have time to pull down this code and try it out locally. Did you get VS Code to behave nicely?

bradbanister:
> Looks nice so far. Do not have time to pull down this code and try it out locally. Did you get VS Code to behave nicely?

It ended up being one of those "reload VS Code and all is well" issues.  🤷‍♂️

lalrg (commented):
It looks good as a spike, it matches the ac for the story but I don't know if having a single project for all the shared code is the best approach or if it would be better to have projects as components of the solution, for example a project for LMS-FileSystem, other for LMS-SyncData and so on.

@bradbanister should I approve it now? or you are going to work on FIZZ-199 in this same branch?

bradbanister:
> It looks good as a spike, it matches the ac for the story but I don't know if having a single project for all the shared code is the best approach or if it would be better to have projects as components of the solution, for example a project for LMS-FileSystem, other for LMS-SyncData and so on.

I'm thinking we start with a single shared package.  We can split later if it becomes too big/starts lacking cohesion.

> 
> @bradbanister should I approve it now? or you are going to work on FIZZ-199 in this same branch?

No need to approve, I'm continuing work here.

lalrg:
> I'm thinking we start with a single shared package. We can split later if it becomes too big/starts lacking cohesion.

Agreed ✔️

.teamcity/_Self/GoogleClassroom/Project.kt
```diff
@@ -21,6 +21,7 @@ object GoogleClassroomProject : Project({
             +:.teamcity => .teamcity
             +:%project.directory% => %project.directory%
             +:eng => eng
+			+:./src/lib => .src/lib
```
stephenfuqua, line 24: look at those extra spaces! 

😆 

I don't really care.

bradbanister: Actually, this is a good catch. They were tabs inserted by Notepad++.  Serves me right for not using VS Code!

bradbanister:
This is now ready for real review.  After pulling down, run a "poetry install" on both GC and Canvas extractors to get the link to the shared library.

src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -145,11 +137,8 @@ def _get_assignments(
     (assignments, udm_assignments_df) = extract_assignments(
         courses, sections_df, sync_db
     )
-    write_multi_csv(
-        udm_assignments_df,
-        datetime.now(),
-        os.path.join(output_directory, ASSIGNMENT_ROOT_DIRECTORY),
-    )
+    logger.info("Writing LMS UDM Assignments to CSV files")
+    write_assignments(udm_assignments_df, datetime.now(), output_directory)
```
stephenfuqua, line 141: Would there be any value to calculating the datetime once and passing the same file to each `write_xyz` statement?

src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -126,11 +121,8 @@ def _get_sections(sync_db: sqlalchemy.engine.base.Engine) -> None:
     logger.info("Extracting Sections from Canvas API")
     (courses, _) = results_store["courses"]
     (sections, udm_sections_df) = extract_sections(courses, sync_db)
-    write_csv(
-        udm_sections_df,
-        datetime.now(),
-        os.path.join(output_directory, SECTIONS_ROOT_DIRECTORY),
-    )
+    logger.info("Writing LMS UDM Sections to CSV file")
```
stephenfuqua, line 124: What do you think about moving these log statements into the shared library? Is there any unwanted impact in terms of  the output logged to the screen/file?

stephenfuqua (approved):
Couple of questions to consider, but I'm not going to hold up the merge for them. @lalrg please also take a look.

@bradbanister @lalrg please merge whenever you're satisfied, without needing any more review from me.

ps. the build is now passing :-).

lalrg (commented):
Everything is looking good, please add src/canvas-extractor/tests/api/test.db to gitignore and remove it from the PR and after that you can merge

src/canvas-extractor/canvas_extractor/__main__.py
lalrg: I don't have any strong opinion on favour or against this change.

PR #134, opened 2021-01-25
[FIZZ-138] System activities

lalrg:


src/canvas-extractor/canvas_extractor/__main__.py
```diff
@@ -205,6 +207,20 @@ def _get_grades() -> None:
     )
 
 
+@catch_exceptions
```
bradbanister, line 210: Just FYI, this will be something to move to the shared library.  Will be very easy to do.

src/canvas-extractor/canvas_extractor/api/authentication_events.py
```diff
@@ -0,0 +1,144 @@
...
+
+    for element in data:
+        if element is not None:
+            element.update(self._extra_attribs)
+            content.append(self._content_class(self._requester, element))
+
+    return content
+
+
+PaginatedList._get_next_page = custom_get_new_page
```
bradbanister, line 61: When you do the monkey patch here, does the new method still work for the other API calls, or is there a required ordering for patching?

bradbanister: I've often seen a patching pattern where you keep a reference to the original method, swap in the new method so it can be used, then later swap back to the original method.

lalrg: This will apply for all calls, I am not sure if this is the right place to leave this code, but the patch is not invassive, it actually is a copy of the original method with an extra validation for login activities.

src/canvas-extractor/canvas_extractor/__main__.py
lalrg: Sounds good

bradbanister (changes requested):
A few issues.  First, you need to version bump the shared library in the pyproject.toml after making a change.  Then there are some type issues:

image

Not sure why mypy isn't complaining, but they are indeed issues (Don't worry about import issues, they'll be fixed after version bump)

bradbanister (approved):
Fixed two additional type issues. Otherwise looks good after lint and run.

PR #137, opened 2021-02-02
[FIZZ-57] Build automation for Canvas Extractor

stephenfuqua:
This pull request includes the File Tester spike of GitHub Actions support. Includes trivial file tester and canvas changes to trigger builds.

Unfortunately there's no templating system that allows re-using the same job definition in multiple workflows, so far as I can tell. Thus there are two different projects with virtually the same workflow.

If this all looks good, then I'll create a ticket for Schoology & Google Classroom. Since they use the same `build.py` they will be trivial and can be handled in the same ticket. Then a second ticket for removing the TeamCity support, although we would want to wait a few weeks before making the full commitment.


lalrg (approved):
Looks good to me

bradbanister (approved):
Ha!  Luis beat me to it.  Looks good.

PR #138, opened 2021-02-05
[FIZZ-201] Publishing Google Classroom on pypi

stephenfuqua:
Many files had to change for this pull request, primarily because of name changes.

* In order to push the GCP extractor to PyPi, the shared library needed to be pushed to PyPi first.
* In doing so, the shared library name was enhanced for clarity, the directory was moved.
* Google Classroom extractor's packages names were also changed, introducing a prefix of "edfi_".
* In both cases, [`twine`](https://twine.readthedocs.io/en/latest/) was installed as a dev dependency for publishing. It is preferable to using `poetry` or `pip` because it communicates over https instead of http.
* Version numbers were bumped to 1.0.0-alpha0 (or as `1.0.0a0` as represented in PyPi), using the `poetry version premajor` command.
* To publish, you must have a username and password on PyPi, and the `ed-fi-alliance` account needs to run the initial publishing step. Once that occurs, that account can grant other accounts "collaborator" access if needed. The best way to provide those credentials to `twine` is through environment variables `TWINE_USERNAME` and `TWINE_PASSWORD`.
* Publish using `build.py`:
   ```bash
   python .\eng\build.py publish google-classroom-extractor
   ```

image


bradbanister (approved):
Looks good, and a manual publish failed successfully (didn't bump version).

One near-term concern is the package setup for the shared library.  It's probably not going to work long-term to need to publish changes to pypi during development just to pick up a file edit to the shared library.

stephenfuqua:
> One near-term concern is the package setup for the shared library. It's probably not going to work long-term to need to publish changes to pypi during development just to pick up a file edit to the shared library.

So far I'm not seeing any good alternatives other than manually copying the shared library into the virtual environment (temporarily).

bradbanister:
> So far I'm not seeing any good alternatives other than manually copying the shared library into the virtual environment (temporarily).

This gives me an interesting idea.  What if you were to create a symlink in the virtual enviroment back to the package...

PR #141, opened 2021-02-08
[FIZZ-214] Update system activity output directory format

lalrg:


src/lib/extractor-shared/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "extractor-shared"
-version = "0.1.9"
+version = "0.1.10"
```
lalrg, line 3: We have to be carefull when bumping versions and merging to main

src/lib/extractor-shared/extractor_shared/csv_generation/write.py
```diff
@@ -262,8 +262,11 @@ def write_system_activities(
     output_directory: str
         is the root output directory
     """
+    output_path = os.path.join(output_directory, SYSTEM_ACTIVITY_ROOT_DIRECTORY)
```
bradbanister, line 265: FYI, I did this as well for the GC/Canvas shared work.

bradbanister (approved):
Looks good

bradbanister:
@lalrg I though this was done, but I don't see the IN/OUT tag logic.

lalrg:
> @lalrg I though this was done, but I don't see the IN/OUT tag logic.

Hmm, it is in the extract facade.
image
This is the output that I get
image

bradbanister:
Confirmed everything is correct.  I was looking at an old extractor run. 🤣

PR #146, opened 2021-02-12
[FIZZ-212] Google and Schoology Build Configs

stephenfuqua:


.github/workflows/file-tester.yml
```diff
@@ -7,12 +7,12 @@ name: File Tester
...
   pull_request:
     paths:
-      - 'utils/file-tester/**.py'
+      - 'utils/file-tester/**'
+      - '!utils/file-tester/**/*.md'
   push:
-    branches:
-      - main
     paths:
-      - 'utils/file-tester/**.py'
```
stephenfuqua, line 15: Note: it seems that you can't combine branches & paths, so if we leave `branches` in then every push to `main` will trigger execution of every job.

lalrg (approved):
Reviewed the generated coverages and all are generated well. Also reviewed the jobs and everything looks fine, just as in TC

PR #147, opened 2021-02-19
[FIZZ-224] Create a package out of file-utils

lalrg:


src/file-utils/pyproject.toml
```diff
@@ -1,16 +1,17 @@
 [tool.poetry]
 name = "edfi-lms-file-utils"
 version = "1.0.0-alpha.0"
-description = "Utilities to facilitate use of the filesystem created by Ed-Fi LMS Extractors"
 homepage = "https://techdocs.ed-fi.org/display/EDFITOOLS/LMS+Toolkit"
 repository = "https://github.com/Ed-Fi-Exchange-OSS/Ed-Fi-X-Fizz"
+description = "Utilities to facilitate use of the filesystem created by Ed-Fi LMS Extractors"
 authors = ["Ed-Fi Alliance, LLC and contributors"]
 license = "Apache-2.0"
```
stephenfuqua, line 8: Needs a readme entry. [Example](https://github.com/Ed-Fi-Exchange-OSS/LMS-Toolkit/blob/cee87531a8b6de16feb703a0d8627163a83eda59/src/canvas-extractor/pyproject.toml)

.github/workflows/file-ultils.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: python ./eng/build.py coverage:html ../src/file-utils
+
+      - name: Archive code coverage results
+        uses: actions/upload-artifact@e448a9b857ee2131e752b06002bf0e093c65e571
+        with:
+          name: code-coverage-report
+          path: src/file-utils/htmlcov/
+
+      - name: Type Check
+        run: python ./eng/build.py typecheck:xml ../src/file-utils
```
stephenfuqua, line 53:
```suggestion
        run: python ./eng/build.py typecheck:xml file-utils
```

The script automatically looks in the `src` directory. So the only time you need to specify directory is when the package is _not_ in `src`.

.github/workflows/file-ultils.yml
```diff
@@ -0,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+name: File Utils
```
stephenfuqua, line 6:
```suggestion
name: File Utilities Library
```

stephenfuqua (changes requested):

stephenfuqua:
Is this ready to come out of draft?

.github/workflows/file-ultils.yml
```diff
@@ -0,0 +1,56 @@
...
+name: File Utilities Library
+on:
+  pull_request:
+    paths:
+      - 'src/file-utils/**'
+      - '!src/file-utils/**/*.md'
+  push:
+    paths:
+      - 'src/file-utils/**'
+      - '!src/file-utils/**/*.md'
```
stephenfuqua, line 15: @lalrg one problem we have is that we don't actually want to trigger on the push to the _feature_ branch - only on the push into the _main_ branch, if someone bypassed the pull request. So what we really want is to combine paths and branches here. Unfortunately I didn't get it working before. However, [this discussion post](https://github.community/t/combine-path-and-tags-conditionals/17064) claims that it is possible. So I am going to try it out here before merging this pull request.
```suggestion
      - '!src/file-utils/**/*.md'
    branches:
      - main
```

src/notebooks/pyproject.toml
```diff
@@ -11,6 +11,7 @@ matplotlib = "^3.3.2"
 pandas = "^1.1.3"
 ipykernel = "^5.3.4"
 jupyter = "^1.0.0"
+edfi-lms-file-utils = "^1.0.0-alpha.0"
```
stephenfuqua, line 14:
```suggestion
edfi-lms-file-utils = "1.0.0a.1"
```

I am surprised this would work with "alpha" in the name, as the system actually publishes to pypi as "1.0.0a0". Brad pointed out that we should probably hard-code to a specific prerelease instead of using the carat `^` notation, which means "use latest 1.0.x", because we might create different pre-releases at the same time and that could cause integration testing problems.

stephenfuqua: But don't make this change yet, as I will use it to test my workflow change above.

utils/file-tester/pyproject.toml
```diff
@@ -9,6 +9,7 @@ license = "Apache-2.0"
 python = "^3.8"
 pandas = "^1.1.4"
 numpy = "1.19.3"
+edfi-lms-file-utils = "^1.0.0-alpha.0"
```
stephenfuqua, line 12:
```suggestion
edfi-lms-file-utils = "1.0.0a0"

```

stephenfuqua (approved):

PR #148, opened 2021-02-26
[LMS-227] Switch Schoology to use shared sync

bradbanister:


stephenfuqua (changes requested):
One problem detected in QA testing: the `userrole` column in the `LMSUser` file is now blank.

bradbanister:
@stephenfuqua Figured it out. My "bug fix" to the empty users files turned out to be incorrect. The actual issue was a type mismatch between id on role_df and role_id on users_df, so the merge wasn't working. Reverted my "fix" and casting types to match.

stephenfuqua (approved):
Everything looks good now, ready to merge.

PR #149, opened 2021-02-26
[LMS-113] LMS Data Store initialization

stephenfuqua:
**Changes**

* Replaces "FIZZ" with "LMS" or "LMS Toolkit" as appropriate throughout the repository.
* Automates installation of User and staging-User tables in both PostgreSQL and SQL Server. NOTE: I did not use either YoYo or Alembic, two Python libraries for database migrations, because they were too heavyweight for our needs and/or do not have good SQL Server support.
* Modified the command line interface to align with other applications.
* And made a few modifications in preparation for publishing.
* Readme updates.
* Removed the `lms-ds-installer` package, which is no longer needed as a separate utility.

**Problems**

* The upload of a 50,000-student user file crashes. I will create a separate ticket to deal with that.
* I will also create tickets for publishing and for additional refactoring to match current coding and testing standards.


src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/initialize_lms_database.sql
```diff
@@ -0,0 +1,13 @@
+-- SPDX-License-Identifier: Apache-2.0
+-- Licensed to the Ed-Fi Alliance under one or more agreements.
+-- The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+-- See the LICENSE and NOTICES files in the project root for more information.
+
+IF NOT EXISTS (SELECT * FROM sys.schemas WHERE name = N'lms')
+EXEC sys.sp_executesql N'CREATE SCHEMA lms';
+
```
stephenfuqua, line 8: Note: if you try to run this script directly in SQL Management Studio then it might error out, because both statements are treated as a single batch. When the second statement runs, it will think that the `lms` schema doesn't exist yet. The normal solution is to add the word `GO` between statements. However, this won't work with the way that I'm running these files through SQL Alchemy: the files are split using `sqlparse`, which doesn't know about `GO`. Thus for this project the semi-colon separator is *essential*. I should mention this in the readme before finishing this ticket.

src/lms-ds-loader/edfi_lms_ds_loader/migrator.py
```diff
@@ -0,0 +1,124 @@
...
+
+def _read_statements_from_file(full_path: str) -> List[str]:
+    raw_sql: str
+    with open(full_path) as f:
+        raw_sql = f.read()
+
+    return split(raw_sql)
+
+
+def _execute_statements(engine: Engine, statements: List[str]):
```
stephenfuqua, line 36: TODO: refactor this function or get rid of it entirely in favor of `_execute_transaction`.

stephenfuqua: done

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/initialize_lms_database.sql
stephenfuqua: done

lalrg (commented):
I see the readme for lms-ds-loader less detailed than the ones for the extractors, which in some way makes sense, because of this is probably not going to be consumed by a BA...
Should we make a distinction? I mean, even when only the extractors are going to be consumed by BA, this lms-ds-loader directory is in the same folder, so it could be a confussion for the non-developer visitors, who could get to this readme and think that the instructions are not user friendly...

My sugestion would be to separate non-extractor projects in a separated folder, or at least make sure that the general readme specifies that there are some projects intended for BA and other for devs.

stephenfuqua:
> I see the readme for lms-ds-loader less detailed than the ones for the extractors, which in some way makes sense, because of this is probably not going to be consumed by a BA...

While I made some updates to the readme, I didn't try to tackle it completely in this ticket as I was trying to keep the scope of work constrained. It does need continuous improvement as we work on other tickets - which means we should "improvement of readme" to several tickets' acceptance criteria.

> Should we make a distinction? I mean, even when only the extractors are going to be consumed by BA, this lms-ds-loader directory is in the same folder, so it could be a confussion for the non-developer visitors, who could get to this readme and think that the instructions are not user friendly...
> 
> My sugestion would be to separate non-extractor projects in a separated folder, or at least make sure that the general readme specifies that there are some projects intended for BA and other for devs.

I understand where you are coming from, but I don't make any distinction between them. The Jupyter notebooks should be the primary source for helping a BA get up and running. Which reminds me that I have been meaning to create a ticket around that. Thus I would prefer to keep them all together, and continue improving the documentation in the repository and in Tech Docs (a BA - as opposed to a developer - might spend more time in Tech Docs initially).

src/lms-ds-loader/README.md
```diff
@@ -5,22 +5,26 @@ Model (LMS-UDM) into a Learning Management System Data Store (LMS-DS) database.
...
 1. Install required Python packages:
 
    ```bash
    poetry install
    ```
 
-1. Run the two SQL scripts in `src/lms-ds-installer/mssql` in a SQL Server
-   database.
 1. The database account used when running the tool needs to be a member of the
    following roles in the destination database:
```
lalrg, line 30: the sample run of line 43 needs to be updated

lalrg (commented):
I left a comment about a change that should be done in the readme, other things that I see:
1- It is not clear if I should create the database before the first run or if it is created by the script.
2- There's .env support in the code but is not documented
3- Even when someone could run `script.py -h` to get information on the valid parameters, it is different from other readmes where we have a table with detailed information in the readme...

Since most of my concerns are related with UX, I don't know if those should be handled in a separated ticket.

src/lms-ds-loader/README.md
stephenfuqua: Updated it, and made several other readme improvements in the latest commit. I've also added "readme improvement" acceptance criteria to several other tickets so that we can continuously improve on this.

lalrg (approved):
The script is creating the expected tables... It generates some errors when loading data but it will be handled in a separated ticket. It is working as expected at this moment.

PR #151, opened 2021-03-02
[LMS-231] Bug fixes for the LMS DS Loader

stephenfuqua:
* Fixed the main bug when uploading a large file: the DataFrame-to-SQL statement was too large, so I set the chunksize to the largest values that was successful.
* Found an overlooked `deletedat` column that should have been in LMS-113. I did not add a separate migration for this, since it was found within the sprint. Anyone who previously installed the tables using this tool after LMS-113 needs to delete all `lms.*` tables in their database.
* Also fixed a bug in the soft delete process, and added a cautionary note in the readme. The problem occurred when I uploaded users from one LMS and then users from a second LMS. The second operation was erroneously flagging the first LMS's records as deleted.


bradbanister (approved):
Looks good

PR #152, opened 2021-03-03
[LMS-76] Automatic publish for Google Classroom

bradbanister:


bradbanister:
A comment

bradbanister:
This works when triggered by an in-branch push.  See: https://github.com/Ed-Fi-Exchange-OSS/LMS-Toolkit/runs/2018651912

Changed to a manual trigger (workflow_dispatch event), but needs to be merged into main for the shiny green button to be visible.  Please review for all but that, and post merge we can do a final test of the manual trigger.

One question is if there is a known better way to share between this publish workflow file and the test workflow file.  I couldn't find an approach that would work.  Multiple jobs in a file didn't seem to fit the bill, and as far as I can tell one workflow triggering another lacks clear support.

.github/workflows/publish-google-classroom.yml
```diff
@@ -0,0 +1,63 @@
...
+        # very carefully use the tag when testing with act-cli on localhost, commit the hash version
+        # uses: actions/cache@v2
+        uses: actions/cache@0781355a23dac32fd3bac414512f4b903437991a
+        with:
+          path: ~/.cache/pypoetry/virtualenvs
+          key: ${{ runner.os }}-poetry-${{ hashFiles('poetry.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-poetry-
+
+      - name: Run Tests with Coverage
```
stephenfuqua, line 44: I agree that we need to re-run some quality checks before publishing, since it is possible to merge something and ignore an error in the trigger on push to `main`. However, perhaps we can simplify a bit by just running `pytest` without capture code coverage? Running type check also helps avoid production bugs, but I wonder if running the linter is really necessary?

eng/build.py
```diff
@@ -166,17 +167,7 @@ def _run_build():
...
-            "del",
-            "/q",
-            "dist\\*.*"
-        ], exit_immediately=False)
-    else:
-        _run_command([
-            "rm",
-            "dist/*"
-        ], exit_immediately=False)
+    shutil.rmtree('dist', ignore_errors=True)
```
stephenfuqua, line 170: 👍

stephenfuqua (approved):
Approved, but not merging pending @bradbanister's response to a question about streamlining.

stephenfuqua:
> One question is if there is a known better way to share between this publish workflow file and the test workflow file. I couldn't find an approach that would work. Multiple jobs in a file didn't seem to fit the bill, and as far as I can tell one workflow triggering another lacks clear support.

Not that I've been able to find... although we might be able to create a custom Action that is a composite of other things. Hadn't thought about that before. I have commented in a GitHub survey that this is the top missing feature in my view.

.github/workflows/publish-google-classroom.yml
bradbanister: That makes sense.  I'll update.

stephenfuqua (approved):

PR #154, opened 2021-03-04
[LMS-93] Logging and Exception Improvements

stephenfuqua:


bradbanister (approved):
The code looks good, as does execution, so I'm approving. However, I'm getting 36 Pylance type errors on the code base.  A few are from tests passing in invalid types, most of the others are around type information for MssqlLmsOperations.  Handle in this ticket, or create a new one?

stephenfuqua:
> The code looks good, as does execution, so I'm approving. However, I'm getting 36 Pylance type errors on the code base. A few are from tests passing in invalid types, most of the others are around type information for MssqlLmsOperations. Handle in this ticket, or create a new one?

Interesting that mypy doesn't detect those problems, even when I added the `tests` directory to the mypy.ini. Pylance type checking was not yet enabled on this project (oversight), so I'm not sure how you saw those errors? I'm enabling it now and will make some updates based on what I see.

bradbanister:
> Interesting that mypy doesn't detect those problems, even when I added the `tests` directory to the mypy.ini. Pylance type checking was not yet enabled on this project (oversight), so I'm not sure how you saw those errors? I'm enabling it now and will make some updates based on what I see.

Yes, it makes me wonder if our mypy config settings need changing, though I don't actually know if that's a thing. On Pylance, I guess I have Pylance type checking enabled in VS Code at the user level.

PR #155, opened 2021-03-04
[LMS-70] automated publish for Schoology

bradbanister:


bradbanister:
A comment

bradbanister:
A comment

bradbanister:
Test published via a temporary "push" trigger, then switched to workflow_dispatch for push button.  See: https://github.com/Ed-Fi-Exchange-OSS/LMS-Toolkit/actions/workflows/publish-schoology.yml

After merge, test with push button trigger.

.github/workflows/publish-schoology.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Cache Dependencies
+        # very carefully use the tag when testing with act-cli on localhost, commit the hash version
+        # uses: actions/cache@v2
+        uses: actions/cache@0781355a23dac32fd3bac414512f4b903437991a
+        with:
+          path: ~/.cache/pypoetry/virtualenvs
+          key: ${{ runner.os }}-poetry-${{ hashFiles('poetry.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-poetry-
```
stephenfuqua, line 41: [This article](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows) describes how it works. And [this one](https://docs.github.com/en/actions/guides/building-and-testing-python) gets into more detail on Python. What's clear is that I misunderstood something here when I applied the caching _after_ installing dependencies. Looks like it should have actually been _before_ installing dependencies. Let's change this around right now and look at the timing in the actual actions, to see if this actually improves the build time.

```suggestion
      - name: Cache Dependencies
        # very carefully use the tag when testing with act-cli on localhost, but use the hash when committing
        # uses: actions/cache@v2
        uses: actions/cache@0781355a23dac32fd3bac414512f4b903437991a
        with:
          path: ~/.cache/pypoetry/virtualenvs
          key: ${{ runner.os }}-poetry-${{ hashFiles('poetry.lock') }}
          restore-keys: |
            ${{ runner.os }}-poetry-
            
      - name: Install Python Dependencies
        run: python ./eng/build.py install schoology-extractor
```

Side note: I forgot to use caching in the MetaEd build. I need to create a ticket for that.

src/schoology-extractor/.vscode/settings.json
```diff
@@ -103,5 +103,5 @@
         "pytest"
     ],
     "jupyter.jupyterServerType": "local",
-    "python.pythonPath": ".venv\\Scripts\\python.exe"
+    "python.pythonPath": "c:\\work\\fizz\\Ed-Fi-X-Fizz\\src\\schoology-extractor\\.venv\\Scripts\\python.exe"
```
stephenfuqua, line 106:
```suggestion
    "python.pythonPath": ".venv\\Scripts\\python.exe"
```
My VS Code never changes this, wonder why yours does?

src/schoology-extractor/edfi_schoology_extractor/helpers/csv_reader.py
```diff
@@ -1,24 +1,24 @@
-# SPDX-License-Identifier: Apache-2.0
```
stephenfuqua, line 1: I wonder if we should just go ahead and add linefeed style to the editorconfig file to avoid having some files with CRLF and some with LF. Hard to tell if any actual changes occurred here. I can see on my local side that the file was previously in CRLF and is now LF - which I'm happy about.

What do you think about adding `end_of_line = lf` to the editorconfig file? @lalrg any concern?

bradbanister: Sounds good to me

lalrg: Sounds good

stephenfuqua: I'll do that in my pull request and will accept & merge this one.

stephenfuqua (approved):

PR #157, opened 2021-03-08
[LMS-232] Refactor lms-ds-loader for standardization & maintainability

stephenfuqua:


src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -1,477 +1,478 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _default() -> pd.DataFrame:
+    return pd.DataFrame()
+
+
+def _read_csv(file: str, nrows: Union[int, None] = None) -> pd.DataFrame:
+    logger.info(f"Reading file: {file}")
```
stephenfuqua, line 22: Argh, the file endings changed! This line above is the only actual change. When I refactored the loader application to use this library, I realized that I had lost logging about what file is being read. This seemed like the best way to re-introduce that logging.

src/lms-ds-loader/edfi_lms_ds_loader/df_to_db.py
```diff
@@ -0,0 +1,45 @@
+# SPDX-License-Identifier: Apache-2.0
```
stephenfuqua, line 1: Unfortunate that GitHub is not recognizing the file rename and displaying changes - this was formerly `csv_to_sql.py`. The name changed when I used the shared library to read the latest CSV into a DataFrame.

src/lms-ds-loader/edfi_lms_ds_loader/migrator.py
```diff
@@ -5,45 +5,26 @@
...
-        response = function(session)
-
-        session.commit()
-
-        return response
-    except ProgrammingError:
-        session.rollback()
-        raise
-    finally:
-        session.close()
```
stephenfuqua, line 43: I moved this function to `sql_adapter.py`.

src/lms-ds-loader/edfi_lms_ds_loader/migrator.py
```diff
@@ -53,57 +34,45 @@ def _read_statements_from_file(full_path: str) -> List[str]:
...
-            # Ignore MSSQL "GO" statements
-            if statement == "GO":
-                continue
-
-            # Deliberately throwing away all results. Counting on exception handling
-            # if there are any errors, and migration scripts should not be returning
-            # any results.
-            session.execute(statement)  # type: ignore
-
-    _execute_transaction(engine, __callback)
```
stephenfuqua, line 71: Also moved the above function to `sql_adapter.py`. I had hoped that doing so would help me with unit testing, but no luck so far.

src/lms-ds-loader/edfi_lms_ds_loader/migrator.py
```diff
@@ -53,57 +34,45 @@ def _read_statements_from_file(full_path: str) -> List[str]:
...
-def _script_has_been_run(engine: Engine, migration: str) -> bool:
-    def __callback(session: sa_session):
-        statement = f"SELECT 1 FROM lms.migrationjournal WHERE script = '{migration}';"
-        return session.execute(statement).scalar()  # type: ignore
-
+def _script_has_been_run(engine: sa_Engine, migration: str) -> bool:
     try:
-        response = _execute_transaction(engine, __callback)
+        statement = f"SELECT 1 FROM lms.migrationjournal WHERE script = '{migration}';"
+        response = get_int(engine, statement)
```
stephenfuqua, line 44: I tried monkey patching `get_int` but it did not work. Now that I'm reviewing the code, I wonder if it is something about the import statement running before the mock is generated - thus the mock is ignored. I seem to recall something about that when I was working more closely on this project last November or December.

Tried `mocker.patch("edfi_lms_ds_loader.sql_adapter.get_int", 0` and also tried injecting a method instead of a number, and it was always ignored.

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
```diff
@@ -1,234 +1,212 @@
...
-select 1 from lms.stg_{table} as stg
-where t.sourcesystemidentifier = stg.sourcesystemidentifier
-and t.sourcesystem = stg.sourcesystem
-) and deletedat is null
-and t.sourceSystem = '{sourceSystem}'
-""".strip()
-
-        rowcount = self._exec(statement)
-        logger.debug(f"Soft-deleted {rowcount} records in table `{table}`")
+# SPDX-License-Identifier: Apache-2.0
```
stephenfuqua, line 1: 🤬 line ending changes

src/lms-ds-loader/edfi_lms_ds_loader/helpers/argparser.py
```diff
@@ -7,6 +7,8 @@
 from typing import List, Union
 
 from configargparse import ArgParser  # type: ignore
+from sqlalchemy import create_engine as sa_create_engine
+from sqlalchemy.engine import Engine as sa_Engine
 
 from edfi_lms_ds_loader.helpers.constants import DbEngine, LOG_LEVELS
```
lalrg, line 13: We could also move constants to the shared package

src/lms-ds-loader/edfi_lms_ds_loader/helpers/argparser.py
```diff
@@ -153,7 +153,7 @@ def parse_main_arguments(args_in: List[str]) -> MainArguments:
 
     USE_INTEGRATED = "--useintegratedsecurity"
```
lalrg, line 154: I like having constants at the top 🤔

lalrg (approved):
Code is still behaving as expected... The comments that I left are not blocking it from be merged.

src/lms-ds-loader/edfi_lms_ds_loader/migrator.py
stephenfuqua: I think you're right! That sounds very familiar and I think I had learned that last fall, but forgotten about it by now.

src/lms-ds-loader/edfi_lms_ds_loader/__main__.py
```diff
@@ -45,19 +40,23 @@ def _configure_logging(arguments: MainArguments):
         level=arguments.log_level,
```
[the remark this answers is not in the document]
stephenfuqua: That is an interesting idea. Certainly our logging is the same in all applications so far. Better to defer that to another ticket. If we had any other extractors then we should consider making that change at that time.

src/lms-ds-loader/edfi_lms_ds_loader/helpers/argparser.py
stephenfuqua: Generally I do that when the constant is for re-usability. This case is more like when you're in TypeScript and you define a little variable to hold a string, making it easier to incorporate the string into a next step, and you realize that there's no reason for the string to change. So you define it as a `const` instead of a plain variable. I don't think we have that construct in Python, so I used uppercase to signal that this is intended for use as an immutable string.

PR #159, opened 2021-03-09
[LMS-115] Load Section Files into LMS Database

stephenfuqua:
* Creates the Section table and paired staging table.
* Uploads the latest section file into the database.
* Corrects a bug with the User upload.

### DIFF
--- a/src/file-utils/edfi_lms_file_utils/file_reader.py
+++ b/src/file-utils/edfi_lms_file_utils/file_reader.py
@@ -4,9 +4,9 @@
 # See the LICENSE and NOTICES files in the project root for more information.
 
 import logging
-from typing import Callable, Optional, Union
+from typing import Callable, Dict, List, Optional
 
-import pandas as pd
+import pandas as pd  # type: ignore
 
 import edfi_lms_file_utils.file_repository as fr
 import edfi_lms_file_utils.constants as c
@@ -18,17 +18,42 @@ def _default() -> pd.DataFrame:
     return pd.DataFrame()
 
 
-def _read_csv(file: str, nrows: Union[int, None] = None) -> pd.DataFrame:
+def _read_csv(
+    file: str,
+    nrows: Optional[int] = None,
+    data_types: Dict[str, str] = dict(),
+    date_cols: List[str] = list(),
+) -> pd.DataFrame:
     logger.info(f"Reading file: {file}")
     if file:
+
+        dates = [
+            "SourceCreateDate",
+            "SourceLastModifiedDate",
+            "CreateDate",
+            "LastModifiedDate",
+            *date_cols,
+        ]
+
+        dtype = {
+            **data_types,
+            "SourceSystemIdentifier": "string",
+            "SourceSystem": "string"
+        }
+
         return pd.read_csv(
-            file, engine="c", parse_dates=True, infer_datetime_format=True, nrows=nrows
+            file,
+            engine="c",
+            parse_dates=dates,
+            infer_datetime_format=True,
+            nrows=nrows,
+            dtype=dtype,
         )
 
     return _default()
 
 
-def get_all_users(base_directory: str, nrows: Union[int, None] = None) -> pd.DataFrame:
+def get_all_users(base_directory: str, nrows: Optional[int] = None) -> pd.DataFrame:
     """
     Reads the most recent users file into a Pandas DataFrame.
 
@@ -46,14 +71,22 @@ def get_all_users(base_directory: str, nrows: Union[int, None] = None) -> pd.Dat
     """
     file = fr.get_users_file(base_directory)
 
+    data_types = {
+        "UserRole": "string",
+        "SISUserIdentifier": "string",
+        "LocalUserIdentifier": "string",
+        "Name": "string",
+        "EmailAddress": "string"
+    }
+
     if file is not None:
-        return _read_csv(file, nrows)
+        return _read_csv(file, nrows, data_types=data_types)
 
     return _default()
 
 
 def get_all_system_activities(
-    base_directory: str, nrows: Union[int, None] = None
+    base_directory: str, nrows: Optional[int] = None
 ) -> pd.DataFrame:
     """
     Reads the most recent system activities files into a Pandas DataFrame.
@@ -84,12 +117,10 @@ def get_all_system_activities(
     df = pd.concat(df_list)
     df.drop_duplicates(inplace=True)
 
-    return df
+    return pd.DataFrame(df)
 
 
-def get_all_sections(
-    base_directory: str, nrows: Union[int, None] = None
-) -> pd.DataFrame:
+def get_all_sections(base_directory: str, nrows: Optional[int] = None) -> pd.DataFrame:
     """
     Reads the most recent sections file into a Pandas DataFrame.
 
@@ -107,14 +138,22 @@ def get_all_sections(
     """
     file = fr.get_sections_file(base_directory)
 
+    data_types = {
+        "SISSectionIdentifier": "string",
+        "Title": "string",
+        "SectionDescription": "string",
+        "Term": "string",
+        "Status": "string",
+    }
+
     if file is not None:
-        return _read_csv(file, nrows)
+        return _read_csv(file, nrows=nrows, data_types=data_types)
 
     return _default()
 
 
 def get_section_associations(
-    base_directory: str, section_id: int, nrows: Union[int, None] = None
+    base_directory: str, section_id: int, nrows: Optional[int] = None
 ) -> pd.DataFrame:
     """
     Reads the most recent section associations file for the given section into
@@ -146,7 +185,7 @@ def _get_data_for_section(
     base_directory: str,
     sections: pd.DataFrame,
     callback: Callable[[str, int, Optional[int]], pd.DataFrame],
-    nrows: Union[int, None] = None,
+    nrows: Optional[int] = None,
 ) -> pd.DataFrame:
     df = pd.DataFrame()
 
@@ -166,7 +205,7 @@ def _get_data_for_section(
 
 
 def get_all_section_associations(
-    base_directory: str, sections: pd.DataFrame, nrows: Union[int, None] = None
+    base_directory: str, sections: pd.DataFrame, nrows: Optional[int] = None
 ) -> pd.DataFrame:
     """
     Reads the most recent section associations files for all given sections into
@@ -192,7 +231,7 @@ def get_all_section_associations(
 
 
 def get_section_activities(
-    base_directory: str, section_id: int, nrows: Union[int, None] = None
+    base_directory: str, section_id: int, nrows: Optional[int] = None
 ) -> pd.DataFrame:
     """
     Reads the most recent section activities file for the given section into
@@ -221,7 +260,7 @@ def get_section_activities(
 
 
 def get_all_section_activities(
-    base_directory: str, sections: pd.DataFrame, nrows: Union[int, None] = None
+    base_directory: str, sections: pd.DataFrame, nrows: Optional[int] = None
 ) -> pd.DataFrame:
     """
     Reads the most recent section activities files for all given sections into
@@ -247,7 +286,7 @@ def get_all_section_activities(
 
 
 def get_assignments(
-    base_directory: str, section_id: int, nrows: Union[int, None] = None
+    base_directory: str, section_id: int, nrows: Optional[int] = None
 ) -> pd.DataFrame:
     """
     Reads the most recent assignments file for the given section into
@@ -276,7 +315,7 @@ def get_assignments(
 
 
 def get_all_assignments(
-    base_directory: str, sections: pd.DataFrame, nrows: Union[int, None] = None
+    base_directory: str, sections: pd.DataFrame, nrows: Optional[int] = None
 ) -> pd.DataFrame:
     """
     Reads the most recent assignments files for all given sections into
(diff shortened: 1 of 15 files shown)


src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -84,12 +117,10 @@ def get_all_system_activities(
     df = pd.concat(df_list)
     df.drop_duplicates(inplace=True)
 
-    return df
+    return pd.DataFrame(df)
```
stephenfuqua, line 120: Surprisingly Pylance was complaining about this, as it inferred that `pd.concat` was returning something other than a DataFrame. I ran a little test in the Python interpreter and confirmed that `pd.DataFrame(df)` works just fine.

src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -18,17 +18,42 @@ def _default() -> pd.DataFrame:
...
+        }
+
         return pd.read_csv(
-            file, engine="c", parse_dates=True, infer_datetime_format=True, nrows=nrows
+            file,
+            engine="c",
+            parse_dates=dates,
+            infer_datetime_format=True,
+            nrows=nrows,
+            dtype=dtype,
```
stephenfuqua, line 50: Here are the changes around data types. Note that the function signature now accepts two optional parameters, for defining columns that should be treated as dates, and for other data types (which will typically be strings, but could be int64 or float). Note that there are some common columns that exist in all cases (I think), so there is no need to retype `SourceSystemIdentifier` each time, for example.

src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -18,17 +18,42 @@ def _default() -> pd.DataFrame:
...
         )
 
     return _default()
 
 
-def get_all_users(base_directory: str, nrows: Union[int, None] = None) -> pd.DataFrame:
+def get_all_users(base_directory: str, nrows: Optional[int] = None) -> pd.DataFrame:
```
stephenfuqua, line 56: This was an opportunistic change for better use of typings.

src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -107,14 +138,22 @@ def get_all_sections(
     """
     file = fr.get_sections_file(base_directory)
 
+    data_types = {
+        "SISSectionIdentifier": "string",
+        "Title": "string",
+        "SectionDescription": "string",
+        "Term": "string",
+        "Status": "string",
+    }
```
stephenfuqua, line 147: An example of mapping data types, with no need to specify the universal data types.

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_section_tables.sql
```diff
@@ -0,0 +1,61 @@
+-- SPDX-License-Identifier: Apache-2.0
+-- Licensed to the Ed-Fi Alliance under one or more agreements.
+-- The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+-- See the LICENSE and NOTICES files in the project root for more information.
+
+CREATE TABLE lms.LMSSection (
```
stephenfuqua, line 6: The MetaEd output has `[]` qualifiers on all of the SQL Server output. I tend to strip that out except where necessary (`[PRIMARY]`, `[Name]`), which makes it easier to copy the MSSQL script to a PGSQL script.

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_section_tables.sql
```diff
@@ -0,0 +1,61 @@
...
+    LMSSectionStatus NVARCHAR(60) NULL,
+    SourceCreateDate DATETIME2(7) NULL,
+    SourceLastModifiedDate DATETIME2(7) NULL,
+    Discriminator NVARCHAR(128) NULL,
+    CreateDate DATETIME2 NOT NULL,
+    LastModifiedDate DATETIME2 NOT NULL,
+    DeletedAt DATETIME2(7) NULL,
+    CONSTRAINT LMSSection_PK PRIMARY KEY CLUSTERED (
+        LMSSectionIdentifier ASC
+    ) WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
```
stephenfuqua, line 23: All of this stuff on line 23 ends up being cut out when moving to PostgreSQL.

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_section_tables.sql
```diff
@@ -0,0 +1,61 @@
+-- SPDX-License-Identifier: Apache-2.0
+-- Licensed to the Ed-Fi Alliance under one or more agreements.
+-- The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+-- See the LICENSE and NOTICES files in the project root for more information.
+
+CREATE TABLE lms.LMSSection (
+    LMSSectionIdentifier INT NOT NULL IDENTITY,
```
stephenfuqua, line 7: For PostgreSQL, `LMSSectionIdentifier SERIAL` if I recall correctly.

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_section_tables.sql
```diff
@@ -0,0 +1,61 @@
...
+    SourceSystemIdentifier NVARCHAR(255) NOT NULL,
```
stephenfuqua, line 8: For PostgreSQL, `VARCHAR` supports Unicode, so `NVARCHAR` --> `VARCHAR`.

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_section_tables.sql
```diff
@@ -0,0 +1,61 @@
...
+    CONSTRAINT LMSSection_PK PRIMARY KEY CLUSTERED (
+        LMSSectionIdentifier ASC
+    ) WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
+) ON [PRIMARY];
+
+ALTER TABLE lms.LMSSection ADD CONSTRAINT LMSSection_DF_CreateDate DEFAULT (getdate()) FOR CreateDate;
+ALTER TABLE lms.LMSSection ADD CONSTRAINT LMSSection_DF_LastModifiedDate DEFAULT (getdate()) FOR LastModifiedDate;
+ALTER TABLE lms.LMSSection ADD CONSTRAINT LMSSection_UK_SourceSystem UNIQUE (SourceSystemIdentifier, SourceSystem);
+
+EXEC sys.sp_addextendedproperty @name=N'MS_Description', @value=N'An organized grouping of course content and users over a period of time for the purpose of providing instruction.', @level0type=N'SCHEMA', @level0name=N'lms', @level1type=N'TABLE', @level1name=N'LMSSection';
```
stephenfuqua, line 30: I don't remember the syntax for these. But we should be able to copy these straight from the MetaEd output.

src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -18,17 +18,42 @@ def _default() -> pd.DataFrame:
...
+            "CreateDate",
+            "LastModifiedDate",
+            *date_cols,
+        ]
+
+        dtype = {
+            **data_types,
+            "SourceSystemIdentifier": "string",
+            "SourceSystem": "string"
+        }
```
lalrg, line 42:
```suggestion
        dtype = {
            "SourceSystemIdentifier": "string",
            "SourceSystem": "string",
            **data_types,
        }
```
In case of having to override types for sourcesystemidentifier, you can use the same data_types parameter to do so if you put it on this way.

stephenfuqua: That's a nice idea, will take it.

src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -18,17 +18,42 @@ def _default() -> pd.DataFrame:
     return pd.DataFrame()
 
 
-def _read_csv(file: str, nrows: Union[int, None] = None) -> pd.DataFrame:
+def _read_csv(
+    file: str,
+    nrows: Optional[int] = None,
+    data_types: Dict[str, str] = dict(),
+    date_cols: List[str] = list(),
```
lalrg, line 25:
```suggestion
    specific_date_cols: List[str] = list(),
```
It'd be helpful to somehow specify that there're some common columns already defined... It cold be renaming the parameter, or adding docstrings(even when it is a privated function)

stephenfuqua: Renaming to `extra_date_columns` seems appropriate, along with a docstring. I'll make that change.

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
```diff
@@ -40,7 +39,6 @@ def run_loader(arguments: MainArguments) -> None:
     db_adapter = arguments.get_db_operations_adapter()
 
     _load_users(csv_path, db_adapter)
```
lalrg, line 41: Similar to what we do when extracting data, it could be helpful to handle exceptions here, so in case a non-critical resource load fails, you can continue with the others and log an error message without breaking the load.

stephenfuqua: That's an interesting idea. 

Currently, let's say that there's a failure when sections loads. The users are already loaded at that point - and we don't wrap a transaction around the entire process, so the users won't roll back.

Let's say that system activities comes after sections in this class. What you're proposing is that we go ahead and load the system activities even though the sections failed. This seems consistent with the fact that we already loaded the users. So I like it. Great suggestion!

stephenfuqua: I ended up writing some unit tests on the `loader_facade` so that I could prove that the try/catch is behaving as expected (which of course is obvious when you look at it, but now the tests also serve as documentation that this is all on purpose). What you suggested about why I was having trouble with monkey patching was true, so you'll see some changes to the import statements on the `loader_facade.py` file.

To test this, I ran the application on the google output. Then changed the name of the `LMSSection` table temporarily in my database, and ran again. Confirmed that I get the desired final logging of the exception, the final warning message, and that the exit code is 1 instead of 0.

image

lalrg: 😎

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
```diff
@@ -12,30 +12,37 @@
...
+from edfi_lms_ds_loader import migrator
 from edfi_lms_file_utils import file_reader
-from edfi_lms_ds_loader.df_to_db import upload_file
+from edfi_lms_ds_loader import df_to_db
 from edfi_lms_ds_loader.mssql_lms_operations import MssqlLmsOperations
 
 logger = logging.getLogger(__name__)
 
 
 def _load_users(csv_path: str, db_adapter: MssqlLmsOperations) -> None:
```
lalrg, line 23:
```suggestion
@catch_exceptions
def _load_users(csv_path: str, db_adapter: MssqlLmsOperations) -> None:
        users = file_reader.get_all_users(csv_path)
        df_to_db.upload_file(db_adapter, users, Table.USER)
```

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
```diff
@@ -12,30 +12,37 @@
...
-    users = file_reader.get_all_users(csv_path)
-    upload_file(db_adapter, users, Table.USER)
+    try:
+        users = file_reader.get_all_users(csv_path)
+        df_to_db.upload_file(db_adapter, users, Table.USER)
+    except BaseException as e:
+        logger.exception(e)
 
 
 def _load_sections(csv_path: str, db_adapter: MssqlLmsOperations) -> None:
```
lalrg, line 31: Same requested change using decorator here

lalrg (changes requested):

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
stephenfuqua: I will try it out...

lalrg: the [@catch_exceptions](https://github.com/Ed-Fi-Exchange-OSS/LMS-Toolkit/blob/main/src/extractor-lib/edfi_lms_extractor_lib/helpers/decorators.py) will log the message and return true or false depending on the resolution of the callable.

stephenfuqua: For a long time I thought it was failing in a surprising way, but I finally figured out that I was making a pytest mistake:

Wrong
```
           def __raise(csv_path):
                raise Exception("bad things")

            mocker.patch("edfi_lms_file_utils.file_reader.get_all_users", return_value=__raise)
```

Right:
```
           def __raise(csv_path):
                raise Exception("bad things")

            mocker.patch("edfi_lms_file_utils.file_reader.get_all_users", side_effect=__raise)
```

(use of `side_effect` instead of `return_value`)

lalrg (approved):
I love to see the different pieces of code that we have developed working all together

PR #161, opened 2021-03-12
[LMS-114] Do not reprocess files

lalrg:


src/file-utils/edfi_lms_file_utils/constants.py
```diff
@@ -1,18 +1,38 @@
...
+LMS_SECTION_SOURCE_SYSTEM_IDENTIFIER = "LMSSectionSourceSystemIdentifier"
+
+# Data types for dataframes
+SECTIONS_DATA_TYPES = {
+    "SISSectionIdentifier": "string",
+    "Title": "string",
+    "SectionDescription": "string",
+    "Term": "string",
+    "Status": "string",
+}
```
stephenfuqua, line 30: Here's why I didn't place them here in the first place: these constants aren't intended for re-use in multiple places. Since they're only needed in the `file_reader.py` module, it is safe to leave them there.

If we have a change in the data types... this constants file might actually be easier to remember where to make the updates. When leaving constants inline (for example, in `file_reader`), there is sometimes the advantage that it is easier to understand the code while reading it because you can see the constant right there. In this case though, it won't make much of a difference I think. Someone reading `file_reader` does not need to know, immediately, about the details of the data types. So it is ok for the reader to just see `data_types = c.USERS_DATA_TYPES` and then they can quickly get to that constant definition if they actually care about the details.

 In general my habit is the original pattern - to extract constants to a separate file when they will be re-used. For this case, I think I'm convinced that it is actually a little better to make the change that you have done.

src/file-utils/edfi_lms_file_utils/file_repository.py
```diff
@@ -1,83 +1,174 @@
...
+
+    if directory is None or not os.path.exists(directory):
+        return None
+
+    files = [
+        {"path": f.path, "name": f.name, "size": f.stat().st_size}
+        for f in os.scandir(directory)
+        if f.name.endswith(".csv")
+    ]
+    files = sorted(files, key=lambda x: str(x["name"]), reverse=False)
```
stephenfuqua, line 23: Do you plan to refactor this function to call `_get_file_paths`?

src/file-utils/edfi_lms_file_utils/file_repository.py
```diff
@@ -1,83 +1,174 @@
...
+        for f in os.scandir(sys_activities):
+            file = _get_newest_file(f.path)
+
+            if file is not None:
+                files.append(file)
+
+    return files
+
+
+def get_system_activities_paths(base_directory: str) -> List[str]:
```
stephenfuqua, line 85: Your other new functions have "files" in the names, for example `get_sections_file_paths`.

src/file-utils/edfi_lms_file_utils/constants.py
lalrg: What do you think about putting those constants at the beginning of the file_reader.py file? my plan is to reuse them in the same file but in new methods

src/file-utils/edfi_lms_file_utils/file_repository.py
lalrg: I plan to extract the common logic of both into a new method and call it from _get_newest_file and _get_file_paths

src/file-utils/edfi_lms_file_utils/file_repository.py
```diff
@@ -1,83 +1,174 @@
...
+        return files
+
+    if os.path.exists(sys_activities):
+        for f in os.scandir(sys_activities):
+            file = _get_file_paths(f.path)
+
+            if file is not None:
+                files = files + file
+
+    return files
```
stephenfuqua, line 100: What is your plan for consolidating the two functions above, to avoid code duplication?

src/file-utils/edfi_lms_file_utils/file_repository.py
lalrg: good catch! thanks

src/file-utils/edfi_lms_file_utils/constants.py
stephenfuqua: Top of the file or here in this constants class, either is fine with me.

src/file-utils/edfi_lms_file_utils/file_repository.py
lalrg: I know it is something to do before merging, but I haven't thought about it. Probably will think about it when writing unit tests.

src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -76,6 +76,44 @@ def _read_csv(
...
+# resource_name
+    data_types = dict()
+    if resource_name == Resources.USERS:
+        data_types = DataTypes.USERS
+
+    if resource_name == Resources.SECTIONS:
+        data_types = DataTypes.SECTIONS
+
+    _read_csv(full_path, nrows, data_types)
+    return _default()
```
stephenfuqua, line 114: Let's make sure this function is well covered by unit tests, particularly since it has branching conditions.

src/file-utils/edfi_lms_file_utils/file_reader.py
```diff
@@ -218,7 +244,9 @@ def _get_data_for_section(
         )
         return _default()
 
-    for _, section_id in sections[[c.SOURCE_SYSTEM_IDENTIFIER]].itertuples():
+    for _, section_id in sections[
+        [Keys.LMS_SECTION_SOURCE_SYSTEM_IDENTIFIER]
```
stephenfuqua, line 248: This is the wrong key - it should be "SourceSystemIdentifier" here, not "LMSSectionSourceSystemIdentifier". I installed this latest version of the file_utils into the `record_counts.ipynb` notebook and ran into an error on this key.
```suggestion
        [Keys.SOURCE_SYSTEM_IDENTIFIER]
```

stephenfuqua: I'll downgrade to the prior version and that should resolve the problem for me.

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_processed_files_table.sql
```diff
@@ -0,0 +1,21 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Licensed to the Ed-Fi Alliance under one or more agreements.
+ * The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+ * See the LICENSE and NOTICES files in the project root for more information.
+ */
+
+CREATE TABLE lms.ProcessedFiles (
+    FullPath NVARCHAR(255) NOT NULL,
```
stephenfuqua, line 9: I was a little concerned if this is a good idea - having the `FullPath` as the primary key. But I think it is the right decision:

1. It is truly unique, so if not primary key then need a separate unique index.
2. We have no reason to think that this table will ever be a foreign key for another table. If that ever happens, then we should introduce an identity column instead.
3. And we don't anticipate updating existing rows, just writing new ones.

src/file-utils/edfi_lms_file_utils/file_reader.py
lalrg: Fixed

src/file-utils/edfi_lms_file_utils/file_repository.py
lalrg: ✔️

src/file-utils/edfi_lms_file_utils/file_repository.py
lalrg: ✔️

src/file-utils/tests/test_file_repository.py
```diff
@@ -1,379 +1,453 @@
...
+        def it_should_return_None(fs, init_fs):
+            file = get_section_activities_file(BASE_DIRECTORY, SECTION_ID)
+            assert file is None
+
+    def describe_when_getting_system_activities_file():
+        def it_should_return_an_empty_list(fs, init_fs):
+            file = get_system_activities_files(BASE_DIRECTORY)
+            assert len(file) == 0
+
+    def describe_when_getting_system_activities_file_paths():
```
lalrg, line 223: Added this test

src/file-utils/tests/test_file_repository.py
```diff
@@ -1,379 +1,453 @@
...
+    def describe_when_getting_system_activities_files():
+        def it_return_newest_file_from_day_one(fs, init_fs):
+            files = get_system_activities_files(BASE_DIRECTORY)
+            assert SYSTEM_ACTIVITIES_FILE_NEW_1 in files
+
+        def it_return_newest_file_from_day_two(fs, init_fs):
+            files = get_system_activities_files(BASE_DIRECTORY)
+            assert SYSTEM_ACTIVITIES_FILE_NEW_2 in files
+
+    def describe_when_getting_system_activities_file_paths_test():
```
lalrg, line 289: Added this test

src/file-utils/tests/test_file_repository.py
```diff
@@ -1,379 +1,453 @@
...
+        def it_should_return_None(fs, init_fs):
+            file = get_section_activities_file(BASE_DIRECTORY, SECTION_ID)
+            assert file is None
+
+    def describe_when_getting_system_activities_files():
+        def it_should_return_an_empty_list(fs, init_fs):
+            files = get_system_activities_files(BASE_DIRECTORY)
+            assert len(files) == 0
+
+    def describe_when_getting_system_activities_file_paths():
```
lalrg, line 145: Added this test

src/file-utils/tests/test_file_repository.py
```diff
@@ -1,379 +1,453 @@
...
+                fs.create_file(newest, contents="\n\n")
+
+                # Act
+                result = _get_newest_file(directory)
+
+                # Assert
+                assert result is None
+
+
+def describe_when_getting_file_paths():
```
lalrg, line 412: Added this tests

src/file-utils/tests/test_file_reader.py
```diff
@@ -1,299 +1,325 @@
...
+
+    def describe_when_getting_attendance_events():
+        def it_should_return_an_empty_DataFrame(mocker, fixture):
+
+            df = get_all_attendance_events(BASE_DIRECTORY, INPUT_DF)
+
+            assert df.empty
+
+
+def describe_when_reading_file_from_path():
```
lalrg, line 309: Added this tests

src/file-utils/tests/test_file_reader.py
```diff
@@ -1,299 +1,325 @@
...
+
+            assert df.iloc[0][ATTENDANCE_EVENTS] == 1
+
+        def given_empty_section_list_it_should_return_empty_data_frame(mocker, fixture):
+
+            df = get_all_attendance_events(BASE_DIRECTORY, pd.DataFrame())
+
+            assert df.empty
+
+    def describe_when_reading_file_from_path():
```
lalrg, line 210: Added this test

src/file-utils/tests/test_file_reader.py
```diff
@@ -1,299 +1,325 @@
...
+    get_all_section_activities,
+    get_all_system_activities,
+    get_all_assignments,
+    get_all_submissions,
+    get_all_grades,
+    get_all_attendance_events,
+    read_file_from_path,
+)
+from edfi_lms_file_utils.constants import Resources, DataTypes
+from .constants import BASE_DIRECTORY
```
lalrg, line 23: I didn't like the idea of having one test importing constants from another test, so I moved the base directory to a constants file and imported it from both test files

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
```diff
@@ -9,29 +9,81 @@
...
+    resource_name: str,
+    get_all_files_callback: Callable,
+):
+    processed_files = db_adapter.get_processed_files(resource_name)
+    all_files: List[str] = get_all_files_callback()
+    all_files_df = pd.DataFrame(all_files, columns=["FullPath"])
+
+    unprocessed_files = pd.concat([processed_files, all_files_df]).drop_duplicates(
+        keep=False
+    )
```
stephenfuqua, line 40: This seems really weird. Why would `unprocessed_files` be created by combining the list of _already processed files_ with the list of all files available? I think I see the answer because of the dropping of duplicates. But it seems really weird reading this way, doesn't it?

https://stackoverflow.com/a/61143201/30384 offers another approach. I'm not convinced that it is any better though!

Seems like these are just lists. Perhaps it would make more sense not to use DataFrames at all.

```python
processed_files: Set[str] = db_adapter.get_processed_files(resource_name)
all_files: Set[str = get_all_files_callback()

return all_files - processed_files
```

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
```diff
@@ -9,29 +9,81 @@
...
+
+    return unprocessed_files
+
+
+def _upload_files_from_paths(
+    db_adapter: MssqlLmsOperations,
+    files: pd.DataFrame,
+    table_name: str,
+    resource_name: str,
+):
```
stephenfuqua, line 51: Return type?

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
```diff
@@ -47,6 +47,11 @@ def __callback(session: sa_Session) -> sa_Result:
 
         return 0
 
+    def _query(self, query: str) -> pd.DataFrame:
```
stephenfuqua, line 50: This is not being re-used. Unless you _know_ that it will be reused, let's recombine it with `get_processed_files`.

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
```diff
@@ -210,3 +215,15 @@ def soft_delete_from_production(self, table: str, sourceSystem: str) -> None:
 
         rowcount = self._exec(statement)
         logger.debug(f"Soft-deleted {rowcount} records in table `{table}`")
+
+    def get_processed_files(self, resource_name: str) -> pd.DataFrame:
+        query = f"""
+SELECT FullPath FROM lms.ProcessedFiles
+WHERE ResourceName = '{resource_name}'""".strip()
+        return self._query(query)
```
stephenfuqua, line 223: Using the DataFrame on line 53 was a nice convenience to avoid writing ORM code and to avoid having to manually iterate over rows and build a list. But in the end it will be easier to deal with the list - or in fact as mentioned above, a set - of file names

```suggestion
    def get_processed_files(self, resource_name: str) -> Set[str]:
        query = f"""
SELECT FullPath FROM lms.ProcessedFiles
WHERE ResourceName = '{resource_name}'""".strip()

        df = pd.read_sql_query(query, self.engine)

       return set(df["FullPath"])
```

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
```diff
@@ -210,3 +215,15 @@ def soft_delete_from_production(self, table: str, sourceSystem: str) -> None:
...
+
+    def add_processed_file(self, path: str, resource_name: str, rows: int):
+        statement = f"""
+INSERT INTO lms.ProcessedFiles(FullPath, ResourceName, NumberOfRows)
+VALUES ('{path}','{resource_name}', {rows})""".strip()
+        return self._exec(statement)
```
stephenfuqua, line 229: I believe this code is safe from SQL injection attacks since we're not taking in any user input. However, it would be better to use [SQL Alchemy's parameter injection method](https://security.openstack.org/guidelines/dg_parameterize-database-queries.html) instead of directly putting the string values into the statement as was done here.

src/file-utils/edfi_lms_file_utils/constants.py
```diff
@@ -1,18 +1,43 @@
...
+
+
+# Keys
+class Keys:
+    LMS_SECTION_SOURCE_SYSTEM_IDENTIFIER = "LMSSectionSourceSystemIdentifier"
+    SOURCE_SYSTEM_IDENTIFIER = "SourceSystemIdentifier"
+
+
+# Data types for dataframes
+
```
stephenfuqua, line 28: I put in a lot of comments while this was in draft. Now that it is out of draft, I need one more comment in order to have an official review.
```suggestion
```

stephenfuqua (dismissed):

src/file-utils/edfi_lms_file_utils/file_repository.py
```diff
@@ -5,61 +5,85 @@
...
-        return None
+@dataclass
+class FileInfo:
+    path: str
+    name: str
+    size: int
+
+    def has_contents(self) -> bool:
+        """ "
+            Determines if a file has contents beyond a couple of line breaks; this situation occurs when
```
stephenfuqua, line 22: There's an extra quotation mark and something happened with the spacing.
```suggestion
        """ 
        Determines if a file has contents beyond a couple of line breaks; this situation occurs when
```

src/file-utils/tests/test_file_reader.py
```diff
@@ -16,10 +17,10 @@
     get_all_assignments,
     get_all_submissions,
     get_all_grades,
-    get_all_attendance_events,
-    read_file_from_path,
+    get_all_attendance_events, read_assignments_file, read_attendance_events_file, read_grades_file, read_section_activities_file, read_section_associations_file, read_sections_file, read_submissions_file, read_system_activities_file,
```
stephenfuqua, line 20: Please reformat the file with `black` to wrap this line.

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
lalrg: set does not support that operation, will use a list instead for the return type of get_all_files_callback

bradbanister (approved):
Run manual tests per LMS-114, everything looks good.  Made a small commit to fix a typing issue.

PR #162, opened 2021-03-13
[LMS-228] Performance testing; copyright update

stephenfuqua:
What's changed?

* Opportunistically, I made some copyright updates that just needed to happen.
* Changed the record count notebook to use the published utilities.
* Updated the sample data generation scripts so that I could easily generate more assignments.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -172,30 +172,3 @@
       defend, and hold each Contributor harmless for any liability
       incurred by, or claims asserted against, such Contributor by reason
       of your accepting any such warranty or additional liability.
-
-   END OF TERMS AND CONDITIONS
-
-   APPENDIX: How to apply the Apache License to your work.
-
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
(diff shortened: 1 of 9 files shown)


LICENSE
```diff
@@ -172,30 +172,3 @@
       defend, and hold each Contributor harmless for any liability
       incurred by, or claims asserted against, such Contributor by reason
       of your accepting any such warranty or additional liability.
-
```
stephenfuqua, line 175: This section is not part of the license and should not have been included.

src/schoology-extractor/tests/data_generation/__main__.py
```diff
@@ -17,6 +17,7 @@
 )
 from tests.data_generation.assignments import (
     generate_and_load_assignments,
+    generate_extra_assignments_without_enrollments,
```
stephenfuqua, line 20: Its not being used, and thus causing a style check fail.

```suggestion
   # generate_extra_assignments_without_enrollments,
```

lalrg (approved):
Overal code looks good and I am happy with the information that we got fro the results of the performance test.

PR #163, opened 2021-03-18
[LMS-117] Upload Assignments

stephenfuqua:
This pull request ended up being far larger than we could have imagined, primarily due to the need to split the Assignments file data into two tables: `lms.Assignment` and `lms.AssignmentSubmissionTypes`. 

There were differences between the three extractors in how they populated the `SubmissionType` column, so testing the output from all three is very important. In addition, I found that the `AssignmentCategory` is not used by some extractors and therefore needs to be nullable. We might want to go back and review the difference between these two columns. However, I will do that separately - out of scope for this ticket.


docs/supporting-artifacts/0020-LMS-Tables.sql
```diff
@@ -1,16 +1,16 @@
-/*
- * NOTE: the tables below have been modified from the output generated by
- * MetaEd: the primary key columns have been set to IDENTITY.
- */
```
stephenfuqua, line 4: At this point I prefer putting the raw MetaEd output into this `supporting-artifact` directory, and then only modify them when copying over into the LMS DS Loader utility.

Also there was a recent update to MetaEd to support putting the file headers in the output when run with "Alliance Mode" enabled. This is very appropriate for these files.

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
```diff
@@ -22,17 +22,37 @@
...
+# failure. For example: users upload fails. Then section would work, but most of
+# the child-objects of section have a dependency on users. If users succeeds,
+# but section fails, then there's little value to having the users and nothing
+# else. When we get that far down, better to keep the pattern of not catching
+# exceptions throughout then to only catch exceptions on "lower value" targets
+# like grades or attendance.
+
+# TODO: consider refactoring for transaction management, so that operations can
+# be rolled back. SQL Alchemy transactions can be used with DataFrame, so this
+# should be feasible.
```
stephenfuqua, line 36:
```suggestion
# TODO: consider refactoring for transaction management, so that operations can
# be rolled back. SQL Alchemy transactions can be used with DataFrame, so this
# should be feasible. https://tracker.ed-fi.org/browse/LMS-244
```

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
```diff
@@ -60,7 +62,7 @@ def truncate_staging_table(self, table: str) -> None:
         assert table.strip() != "", "Argument `table` cannot be whitespace"
 
         # Note: for postgresql we'll want `TRUNCATE TABLE {staging} RESTART IDENTITY`
-        self._exec(f"truncate table lms.[stg_{table}];")
+        self._exec(f"TRUNCATE TABLE lms.stg_{table};")
```
stephenfuqua, line 65: I changed the SQL statements to use upper case. Neither SQL Server nor PostgreSQL "cares" about the case, but I find it easier to read because it is simple to distinguish the SQL language components from the data structure / names.

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
```diff
@@ -136,21 +138,91 @@ def insert_new_records_to_production(self, table: str, columns: List[str]) -> No
...
 
         statement = f"""
-insert into lms.[{table}] ( {column_string} )
-select {column_string}
-from lms.stg_{table} as stg
-where not exists (
-  select 1 from lms.[{table}]
-  where sourcesystemidentifier = stg.sourcesystemidentifier
-  and sourcesystem = stg.sourcesystem
+INSERT INTO
```
stephenfuqua, line 144: I reformatted these statements into a form I have been using with Analytics Middle Tier, which I think it generally very easy to read. The above was previously easy to read, but I think this expanded format is better - especially when you after the `{column_string}` substitution, and you look at the actual SQL statement that is sent to the server.

src/lms-ds-loader/edfi_lms_ds_loader/sql_adapter.py
```diff
@@ -5,17 +5,18 @@
...
 from sqlalchemy.exc import ProgrammingError
 
+T = TypeVar('T')
+
 
 def execute_transaction(
-    engine: sa_Engine, function: Callable[[sa_Session], Optional[sa_Result]]
-) -> Optional[sa_Result]:
+    engine: sa_Engine, function: Callable[[sa_Session], T]
+) -> T:
```
stephenfuqua, line 19: I figured out how to use generics here! That fixed some confusing type issues I  had before.

bradbanister: Very cool!

bradbanister (approved):
Looks good. Tested on all three extractors. Found issue with Canvas extractor assignment uniqueness, will be addressed in LMS-245.

PR #168, opened 2021-03-26
[LMS-118] Load AssignmentSubmission Files

lalrg:
Still missing the upload to DB method to handle foreign keys to sections and users and tests.


src/lms-ds-loader/tests/test_df_to_db.py
```diff
@@ -308,3 +308,66 @@ def it_soft_deletes_from_production_table(
...
+
+        # Act
+        df_to_db.upload_assignment_submissions(adapter_mock, df)
+
+        return adapter_mock, df, insert_mock
+
+    def it_disables_the_natural_key_index(when_uploading_assignment_submissions) -> None:
+        adapter_mock, _, _ = when_uploading_assignment_submissions
+        assert adapter_mock.disable_staging_natural_key_index.call_args_list == [
+            call(Table.ASSIGNMENT_SUBMISSION)
```
stephenfuqua, line 334: I hadn't come across `unittest.mock.call` before. Looks handy.

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_assignment_submission_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+CREATE TABLE lms.AssignmentSubmission (
+    AssignmentSubmissionIdentifier INT NOT NULL IDENTITY,
+    SourceSystemIdentifier NVARCHAR(255) NOT NULL,
+    SourceSystem NVARCHAR(255) NOT NULL,
+	AssignmentIdentifier INT NOT NULL,
+	LMSUserIdentifier INT NOT NULL,
+    SubmissionStatus NVARCHAR(255) NULL,
+	SubmissionDateTime DATETIME2 NULL,
+	EarnedPoints decimal NULL,
+	Grade decimal NULL,
```
stephenfuqua, line 15: This `Grade` column data type does not match the UDM, which has varchar(20) for supporting letter grades.

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_assignment_submission_tables.sql
```diff
@@ -0,0 +1,72 @@
...
+CREATE TABLE lms.stg_AssignmentSubmission (
+    StagingId INT NOT NULL IDENTITY,
+    SourceSystemIdentifier NVARCHAR(255) NOT NULL,
+    SourceSystem NVARCHAR(255) NOT NULL,
+	AssignmentSourceSystemIdentifier NVARCHAR(255) NOT NULL,
+	LMSUserSourceSystemIdentifier NVARCHAR(255) NOT NULL,
+    SubmissionStatus NVARCHAR(255) NULL,
+	SubmissionDateTime DATETIME2 NULL,
+	EarnedPoints decimal NULL,
+	Grade decimal NULL,
```
stephenfuqua, line 60: Same as above - wrong data type.

stephenfuqua (changes requested):
Requesting a few format updates, and there is an incorrect data type.

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
```diff
@@ -13,7 +13,7 @@
 from typing import Callable, Optional, Set
 from functools import lru_cache
 
-import pandas as pd
+from pandas import DataFrame
```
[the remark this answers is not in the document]
bradbanister: Not a huge deal to me either, but I chose to go this way because we're solely importing Pandas for DataFrame as a type annotation.

src/lms-ds-loader/tests/test_df_to_db.py
bradbanister: Yeah, `call_args_list` seems to be the way to go if you want to be exact about number of calls and parameters of each call.

stephenfuqua (changes requested):
I am unable to load the sample files, getting a datetime error. I manually fixed something and still get the error. I noticed the following in the console error message:

image

That is, the `SubmissionDate` is not being handled as a datetime by Pandas. This might be related to the upload problem.

Which draws my attention to the extra column mappings we were applying in `file_reader.py` in  the `file_utils` library. Can you go ahead and add the data type mappings there? That might solve the problem, and would give us a little more control & consistency with other resource handling.

src/file-utils/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "edfi-lms-file-utils"
-version = "1.0.0-alpha.11"
+version = "1.0.0-beta.11"
```
stephenfuqua, line 3: The name doesn't really matter... just fyi the only reason I had used `beta` while you were out was to avoid confusion with Luis's work. No problem leaving this.

bradbanister: I went to beta because Poetry wasn't seeing my alpha.13 release on PyPi for some reason. It saw my previous bump to alpha.12 just fine, so don't know what the issue was but bumping to beta fixed it.  Probably makes more sense anyway to have it as the proper "latest".

stephenfuqua (approved):

PR #171, opened 2021-04-01
[LMS-122] Load section activities

bradbanister:


stephenfuqua (commented):
Just for consistency, would it make sense to add manual data type / datetime mappings in the shared library?

src/lms-ds-loader/edfi_lms_ds_loader/scripts/mssql/create_section_activity_tables.sql
```diff
@@ -0,0 +1,75 @@
...
+EXEC sys.sp_addextendedproperty @name=N'MS_Description', @value=N'The source system datetime the record was last modified.', @level0type=N'SCHEMA', @level0name=N'lms', @level1type=N'TABLE', @level1name=N'LMSSectionActivity', @level2type=N'COLUMN', @level2name=N'SourceLastModifiedDate';
+EXEC sys.sp_addextendedproperty @name=N'MS_Description', @value=N'The date and time at which a record was detected as no longer available from the source system, and thus should be treated as ''deleted''.', @level0type=N'SCHEMA', @level0name=N'lms', @level1type=N'TABLE', @level1name=N'LMSSectionActivity', @level2type=N'COLUMN', @level2name=N'DeletedAt';
+
+CREATE TABLE lms.stg_LMSSectionActivity (
+    StagingId INT NOT NULL IDENTITY,
+    SourceSystemIdentifier NVARCHAR(255) NOT NULL,
+    SourceSystem NVARCHAR(255) NOT NULL,
+    LMSUserSourceSystemIdentifier NVARCHAR(255) NOT NULL,
+    LMSSectionSourceSystemIdentifier NVARCHAR(255) NOT NULL,
```
stephenfuqua, line 59: I was unable to process the Schoology files, because they have `LMSSectionIdentifier` instead of `LMSSectionSourceSystemIdentifier`. That's a bug in the Schoology extractor. I will hand-modify the Schoology output in order to finish testing, and report the bug separately.

stephenfuqua: Created the bug and put it directly into our sprint, since we seem to have capacity.

stephenfuqua (approved):

stephenfuqua (changes requested):
Accidentally approved instead of requesting changes: the manual data type mapping. Testing worked after the manual modification to Schoology, but I think it will be better to have the explicit column mapping.

bradbanister:
> Accidentally approved instead of requesting changes: the manual data type mapping. Testing worked after the manual modification to Schoology, but I think it will be better to have the explicit column mapping.

Explicit column mapping for section activities should already be there.  I put it in at the same time I did section associations as a PR suggestion since I was working on both at the time.

stephenfuqua (approved):

PR #173, opened 2021-04-05
[LMS-252] Schoology extractor issues with Session Activity

bradbanister:
Currently includes some LMS-237 commits, will fix once LMS-237 is merged to main

### DIFF
--- a/src/schoology-extractor/edfi_schoology_extractor/api/__init__.py
+++ b/src/schoology-extractor/edfi_schoology_extractor/api/__init__.py
@@ -3,4 +3,4 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
-__version__ = '0.1.0'
+__version__ = "0.1.0"
--- a/src/schoology-extractor/edfi_schoology_extractor/api/__init__.py
+++ b/src/schoology-extractor/edfi_schoology_extractor/api/__init__.py
@@ -2,5 +2,3 @@
 # Licensed to the Ed-Fi Alliance under one or more agreements.
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
-
-__version__ = "0.1.0"
--- a/src/schoology-extractor/edfi_schoology_extractor/api/paginated_result.py
+++ b/src/schoology-extractor/edfi_schoology_extractor/api/paginated_result.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
 
 
 class PaginatedResult:
-    '''
+    """
     The PaginatedResult class is bound with information from the response of the Schoology
     api, specifically when it returns a list of items that could be paginated.
 
@@ -41,7 +41,7 @@ class PaginatedResult:
         The URL where you got the response from.
     current_page_items : list
         The list of items for the current page.
-    '''
+    """
 
     def __init__(
         self,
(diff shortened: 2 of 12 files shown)


src/schoology-extractor/edfi_schoology_extractor/api/request_client.py
```diff
@@ -184,7 +184,7 @@ def _check_response(
         self._check_for_success(response, success_status)
 
     @retry(
-        retry_on_exceptions=(ConnectionError, HTTPError, ProtocolError, Timeout),
+        retry_on_exceptions=(ConnectionError, HTTPError, ProtocolError, Timeout, RuntimeError),
```
stephenfuqua, line 187: I wonder about this. I have seen 502 from Schoology before, and I tend to think that complete failure is a good thing: while the service will come up at some point, a 502 probably means they're having severe problems and thus we might want to stop our requests.

If we *don't* handle it here, then we might want to document. Overall, I had not thought about documenting the retry functionality, but for end users this might be very valuable.

bradbanister: Yeah, I'm not 100% on this mostly because "RuntimeError" is such a blanket exception. However, I got 502's several times while testing and re-running immediately went fine, which leads me to think it's just some sporadic issue.  Our default retry doesn't really retry too hard.  With time-limited exponential backoff, I think we retry four times in one minute before giving up.

stephenfuqua: If it fails four times, the whole script will bomb out, right? Rather than moving on to the next resource? If so that seems reasonable to me.

bradbanister: That's correct.  It rethrows the exception at that point.

src/schoology-extractor/edfi_schoology_extractor/api/__init__.py
```diff
@@ -3,4 +3,4 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
-__version__ = '0.1.0'
+__version__ = "0.1.0"
```
stephenfuqua, line 6: Do you remember why we even have this setting here? I don't think we should be using it, as it doesn't actually mean anything.

src/schoology-extractor/edfi_schoology_extractor/api/paginated_result.py
```diff
@@ -52,12 +52,24 @@ def __init__(
...
+        ), "Argument `api_response` should be of type `dict`."
+        assert isinstance(
+            resource_name, str
+        ), "Argument `resource_name` should be of type `str`."
+        assert isinstance(
+            requested_url, str
+        ), "Argument `requested_url` should be of type `str`."
+        assert isinstance(
+            current_page, int
+        ), "Argument `current_page` should be of type `int`."
```
stephenfuqua, line 72: We should be able to get rid of these now, right? Since we're using good typing hygiene?

stephenfuqua (commented):
I have a couple of questions that are not essential. I'm going to test this now before approving.

src/schoology-extractor/edfi_schoology_extractor/api/__init__.py
bradbanister: Agree.  I think it's from before we knew much about Python.

src/schoology-extractor/edfi_schoology_extractor/api/paginated_result.py
bradbanister: Will do.

src/schoology-extractor/edfi_schoology_extractor/api/__init__.py
bradbanister: Also, that edit was from an automated "black" reformat of the whole project via the command line.  The black VS Code integration isn't working for me right now either.

stephenfuqua (approved):

stephenfuqua (changes requested):
Some unit tests are failing. They probably just need to be removed because type assertions were removed.

stephenfuqua (approved):

PR #174, opened 2021-04-05
[LMS-247] Set a value for AssignmentCategory in Canvas Extraction

stephenfuqua:


lalrg (approved):
Everything is looking good!

PR #176, opened 2021-04-12
[LMS-229] Update notebook dependencies

bradbanister:
There were issues found with notebooks and the GC extractor unrelated to Python 3.9, which I'll put in separate tickets.


stephenfuqua (changes requested):
Everything seems fine so far... except that each `pyproject.toml` still has this:

```
[tool.poetry.dependencies]
python = "^3.8"
```

Seems like we should update that...

bradbanister:
> Everything seems fine so far... except that each `pyproject.toml` still has this:
> 
> ```
> [tool.poetry.dependencies]
> python = "^3.8"
> ```
> 
> Seems like we should update that...

I thought about that but wasn't sure we wanted to make that restriction. The ticket was originally to ensure everything ran on 3.9. Do we now only support 3.9?

stephenfuqua:
> I thought about that but wasn't sure we wanted to make that restriction. The ticket was originally to ensure everything ran on 3.9. Do we now only support 3.9?

The readme updates state that 3.9 is required. I think it will be too confusing for us to maintain both 3.8 and 3.9 right now, as we would need to be able to switch back and forth easily to run tests.

stephenfuqua (approved):

PR #179, opened 2021-04-19
[LMS-239] End-to-end demonstration notebooks

stephenfuqua:
Refactoring was  required in each extractor in order to facilitate calling the extractor logic from a notebook instead of from `__main__.py`.


bradbanister (approved):
The notebooks look good.  I'll approve, but consider adding a README.md to this folder.  The install instructions are in the notebook, but you might need something to get a user to that point.

stephenfuqua:
> The notebooks look good. I'll approve, but consider adding a README.md to this folder. The install instructions are in the notebook, but you might need something to get a user to that point.

Yes, that is a good idea. And a link from the main readme. Will add that now before merging.

PR #180, opened 2021-04-21
[LMS-256] Fix soft delete issues for dependent entities

bradbanister:


src/lms-ds-loader/edfi_lms_ds_loader/df_to_db.py
```diff
@@ -51,6 +51,7 @@ def upload_file(
     df: pd.DataFrame,
     table: str,
     db_adapter_insert_method: Callable[[MssqlLmsOperations, str, List[str]], None],
+    db_adapter_delete_method: Callable[[MssqlLmsOperations, str, str], None],
```
stephenfuqua, line 54: Please add to the docstring below.

src/lms-ds-loader/edfi_lms_ds_loader/loader_facade.py
```diff
@@ -132,23 +114,16 @@ def _load_assignments(csv_path: str, db_adapter: MssqlLmsOperations) -> None:
...
-        for section in sections:
-            paths = paths + file_repository.get_assignments_file_paths(
-                formatted_path, section
-            )
+    formatted_path = abspath(csv_path)
+    file_paths: List[str] = []
+    for section in sections:
+        file_paths = file_paths + file_repository.get_assignments_file_paths(
+            formatted_path, section
+        )
```
stephenfuqua, line 122: It is bugging me that I can't recall why this was done with a callback originally. There was a reason for it at the time.

stephenfuqua (commented):
One small documentation request and an implied question. I need to test this still.

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
```diff
@@ -589,6 +633,219 @@ def soft_delete_from_production(self, table: str, source_system: str) -> None:
...
+            u.LMSUserIdentifier
+        FROM
+           lms.LMSUser as u
+        INNER JOIN
+            lms.stg_{table} as stg
+        ON
+            stg.LMSUserSourceSystemIdentifier = u.SourceSystemIdentifier
+        AND
+            stg.SourceSystem = u.SourceSystem
+    )
```
stephenfuqua, line 776: This clause doesn't seem to be appropriate for Attendance Events.

* Assume school is tracking negative attendance
* One day a teacher accidentally marks a student as absent
* --> there's a record for the student in the file
* Next day fixes it
* --> there's no record the student in the file

This clause will ignore the missing record, since there is no replacement record for that student. I've pushed a new commit with updated sample-out data that includes this scenario.

stephenfuqua (changes requested):

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
stephenfuqua: I think this is also true for Section Association and Section Activities. Ultimately I think the soft delete test just needs to ensure that the discovered records belong to the section that was just uploaded, without needing to look at the user identifier at all.

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
```diff
@@ -589,6 +633,219 @@ def soft_delete_from_production(self, table: str, source_system: str) -> None:
...
```
stephenfuqua, line 847: As with the "section-and-user" relationship, I think this check for the user identifier is unnecessary here.

src/lms-ds-loader/edfi_lms_ds_loader/mssql_lms_operations.py
bradbanister: Yes, I think you are right.  It's only important at the file partition level (section).

stephenfuqua (approved):

PR #183, opened 2021-04-24
[LMS-112] Fix un-soft delete bug

bradbanister:
Automated tests require a SQL Server database named "test_integration_lms_toolkit" to exist.  No other manual interventions required.


bradbanister:
It wasn't a bug.  It was bad manual testing.  I overlooked updating the LastModifiedDate on a changed record.

The new automated tests cover updates and soft delete/un-soft delete for User entities, proving it all works. The transaction start and rollback scheme is working great.  

The only wrinkle is in test organization.  In practice, these sorts of tests need "act" and "assert" in phases  (e.g. initial upload, followed by upload with missing user, followed by upload with restored user).  The transaction is in the context of a single test, and anyway it's not possible in pytest to impose an ordering on tests (looked at pytest-ordering plugin, but it's alpha quality). So, our typical arrange/act organization in a fixture followed by fine grained asserts in separate test methods won't work.

I tried organizing by putting each phase of a test in separate non-test methods inside the test, then calling each in order. It works fine, but I'm mildly concerned that a phase could get skipped by mistake since a test author/maintainer needs to remember they need to be explicitly called. Something to keep in mind.

eng/build.py
```diff
@@ -1,207 +1,211 @@
-# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: Don't know why this diff is so terrible.  Just changed all the pytest executions to specify the "tests" directory.

src/lms-ds-loader/README.md
```diff
@@ -57,6 +57,10 @@ poetry run python edfi_lms_ds_loader --server localhost --dbname lms_toolkit --u
 
 ## Developer Notes
 
+### Dev test database
+
+Create an empty SQL Server database named "test_integration_lms_toolkit". This is required to run the SQL Server integration pytests.
+
```
stephenfuqua, line 63:
```suggestion
### Database-Enabled Integration Tests

This package contains SQL Server-integrated tests, in addition to the standard
unit tests. The specialized tests are in the `tests_integration_sql` directory
and they will run automatically if you call `pytest` with no arguments.

Before running the tests, create an empty SQL Server database named
"test_integration_lms_toolkit": `sqlcmd -Q "create database
test_integration_lms_toolkit"`. This is required to run the SQL Server
integration pytests.

To run only the _unit tests_: `poetry run pytest tests`. To run only the
integration tests, `poetry run pytest tests_integration_sql`.

#### Writing New Tests

Please note and follow the existing pattern for database-enabled integration
tests, as this pattern was carefully tuned to support use of database
transactions for rolling back changes and thereby leaving behind a clean
database.

```

src/lms-ds-loader/tests_integration_sql/conftest.py
```diff
@@ -0,0 +1,78 @@
...
+from sqlalchemy import create_engine
+from sqlalchemy.engine.base import Engine, Connection, Transaction
+
+from edfi_lms_ds_loader.migrator import migrate
+from edfi_lms_ds_loader.mssql_lms_operations import MssqlLmsOperations
+
+
+def _new_mssql_engine() -> Engine:
+    return create_engine(
+        "mssql+pyodbc://localhost,1433/test_integration_lms_toolkit?driver=ODBC+Driver+17+for+SQL+Server?Trusted_Connection=yes",
```
stephenfuqua, line 19: I'm not worried about it right now, but we should probably find a way to pass in a different server or at least port number, for example if someone wants to use SQL Server only in a container and has it mapped to a different port. Also would need to configure password authentication in that case.

src/lms-ds-loader/tests_integration_sql/conftest.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Fixture that sets up a connection to use, and migrate the tables.
+    Assumes existence of local SQLServer DB named 'test_integration_lms_toolkit'
+    """
+    engine = _new_mssql_engine()
+    migrate(engine)
+
+    connection = engine.connect()
+    yield connection
+    connection.close()
```
stephenfuqua, line 34: Nice usage of `yield`. I've never really figured out when to use that before. I've wanted this exact pattern before and didn't think of it. Instead I tend to pass a callback into this function, which can be called at line 33. I like this `yield` pattern better, since it is easier to reason about program flow without callbacks.

@lalrg please also take a look at this.

src/lms-ds-loader/tests_integration_sql/conftest.py
```diff
@@ -0,0 +1,78 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+from typing import Tuple
```
stephenfuqua, line 6:
```suggestion
from typing import Iterable, Tuple
```

src/lms-ds-loader/tests_integration_sql/conftest.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def _new_mssql_engine() -> Engine:
+    return create_engine(
+        "mssql+pyodbc://localhost,1433/test_integration_lms_toolkit?driver=ODBC+Driver+17+for+SQL+Server?Trusted_Connection=yes",
+    )
+
+
+@pytest.fixture(scope="session")
+def mssql_connection():
```
stephenfuqua, line 24:
```suggestion
def mssql_connection() -> Iterable[Connection]:
```

stephenfuqua (approved):

PR #186, opened 2021-04-29
[LMS-262] Additional exception handling

stephenfuqua:


lalrg (changes requested):
Now that I am testing it, I got some errors, and by reviewing the code, I saw the implementation of the catch_exceptions is not the same as used in another extractor, here you have an example from the canvas extractor:
```python3

@catch_exceptions
def _get_sections(
    arguments: MainArguments, sync_db: sqlalchemy.engine.base.Engine
) -> None:
    (courses, _) = results_store["courses"]
    (sections, udm_sections_df, all_section_ids) = extract_sections(courses, sync_db)
    logger.info("Writing LMS UDM Sections to CSV file")
    write_sections(udm_sections_df, datetime.now(), arguments.output_directory)
    results_store["sections"] = (sections, udm_sections_df, all_section_ids)
```

It does the extraction and handles all the extract processes and then writes to a shared variable to be consumed by other methods. 
The reason for that is that the `catch_exceptions` decorator modifies the response and returns true when the call was successful and false when it failed... You can see why it was implemented on that way, in this piece of code:
```python3
    succeeded = _get_courses(
        arguments,
        get_canvas_api(arguments.base_url, arguments.access_token),
        sync_db
    )
    if not succeeded:
        _break_execution("Courses")

    succeeded = _get_sections(arguments, sync_db)
    if not succeeded:
        _break_execution("Sections")
```

stephenfuqua:
> Now that I am testing it, I got some errors, and by reviewing the code, I saw the implementation of the catch_exceptions is not the same as used in another extractor, here you have an example from the canvas extractor:

Can you be more specific here? I don't understand what change you are suggesting.

PR #192, opened 2021-05-08
[LMS-271] Prevent canvasapi logger from being set lower than warning level

bradbanister:


bradbanister:
Confirmed this only affects the Canvas extractor.

src/canvas-extractor/edfi_canvas_extractor/__main__.py
```diff
@@ -27,6 +27,11 @@ def _configure_logging(arguments: arg_parser.MainArguments) -> None:
         level=arguments.log_level,
     )
 
+    # Canvas API logging should never be finer-grained than warning, to avoid PII in logs
+    canvasapi_logger = logging.getLogger("canvasapi")
+    if canvasapi_logger.getEffectiveLevel() < logging.WARN:
+        canvasapi_logger.setLevel(logging.WARN)
```
stephenfuqua, line 33: I think this is the appropriate place for this. However, we probably need to do the same in the canvas-e2e demonstration notebook to make sure that those who want to run from another python script will know to make this change.

stephenfuqua: I will take care of this as part of the QA cycle.

stephenfuqua (approved):

PR #193, opened 2021-05-18
[LMS-264]

lalrg:


src/canvas-extractor/edfi_canvas_extractor/extract_facade.py
```diff
@@ -161,6 +254,17 @@ def _get_grades(arguments: MainArguments) -> None:
     )
 
 
+@retry(
+    retry_on_exceptions=(
+        IOError,
+        CanvasException,
+        RequestException,
+        socket.timeout,
+        socket.error,
```
stephenfuqua, line 263: Don't forget about adding `ConnectionAbortedError`. Actually, that is a sub-class of `ConnectionError`. These are the base Python types, not library-specific types. In Schoology we are retrying on `ConnectionError`, so let's add that one here too.

lalrg: ✔️

src/canvas-extractor/edfi_canvas_extractor/config.py
```diff
@@ -52,3 +55,21 @@ def get_canvas_api(canvas_base_url: str, canvas_access_token: str) -> Canvas:
         a new CanvasAPI object
     """
     return Canvas(canvas_base_url, canvas_access_token)
+
+
+MAX_TOTAL_CALLS = 4
+RETRY_WINDOW_AFTER_FIRST_CALL_IN_SECONDS = 600
```
lalrg, line 61: What the retry decorator does is to validate:
1. If the max total tries calls have not been reached.
1. Regardless of the number of tries, it validates if the total time of execution for the specific function has taken longer than this parameter, so sometimes if it does not have enough time, it won't retry to execute it.

stephenfuqua (commented):
Looks good. I'm going to test before approving & merging.

stephenfuqua (commented):
Test results are positive, therefore merging.

stephenfuqua (approved):

PR #194, opened 2021-05-19
[LMS-267] Scripts for deleting hardcoded ranges of Schoology and Canvas

bradbanister:
sandbox data

### DIFF
--- a/src/canvas-extractor/tests/data_generation/data_remover/__main__.py
+++ b/src/canvas-extractor/tests/data_generation/data_remover/__main__.py
@@ -1,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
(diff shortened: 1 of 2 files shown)


src/canvas-extractor/tests/data_generation/data_remover/__main__.py
```diff
@@ -0,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: Including the Canvas script work-in-progress here. Referenced in the split ticket.

src/schoology-extractor/data_generation/data_remover/__main__.py
```diff
@@ -0,0 +1,55 @@
+# SPDX-License-Identifier: Apache-2.0
```
bradbanister, line 1: This is the "final" version of the script to delete users and sections. It can be run/re-run without any further effect on the Schoology sandbox.

stephenfuqua (approved):
Code looks good, and the download completes quickly now.

PR #196, opened 2021-05-25
[LMS-272] Canvas extractor performance improvements

bradbanister:


bradbanister:
This change is from getting submissions for each individual assignment per section to a single multi-submission fetch for each section. Mario found discrepancies between the two methods, and I did too but they are not the same (!)

The new fetch method does indeed get submissions for students whose course enrollment is completed.  That is what Mario saw, and that is a win.

However, the new fetch method definitely does not include submissions for unpublished assignments, and the original fetch method did include them.  This is the opposite of what Mario saw.  However, the behavior in the way I'm seeing it makes more sense to me.  It would be weird for a "get all submissions for this assignment" to return nothing if the assignment is unpublished.

stephenfuqua (approved):
Code looks fine to me, but I have not done any testing. Therefore I'm not going to merge. Also giving @lalrg a chance to review.

lalrg (approved):
Code looks good, I did a comparison between the results with and without the changes of this PR, and the information extracted is consistent between them.

PR #210, opened 2021-06-22
[LMS-327] configuration of integration test db connection via pytest command line args

bradbanister:


src/lms-harmonizer/tests_integration_sql/README.md
```diff
@@ -0,0 +1,38 @@
+# Integration test setup
```
lalrg, line 1: I'd add a link to this page from the main readme of the lms_harmonizer

lalrg: In the Dev Operations section perhaps.

lalrg (approved):
I asked for a small change but it does not need a second revision. I approve it so after the small change in the readme @bradbanister can directly merge.

Code looks good and behaves as expected

PR #212, opened 2021-06-25
[LMS-329] Order migratation scripts with numeric prefix

bradbanister:


bradbanister:
Did this real quick to make db creation easier for LMS-330

stephenfuqua (changes requested):
Can you add a small script to convert an existing database? That is, update the records in the MigrationJournal so that it is possible to run this on a database that already has all of the tables.

stephenfuqua:
> Can you add a small script to convert an existing database? That is, update the records in the MigrationJournal so that it is possible to run this on a database that already has all of the tables.

I've done this, adding a note to the readme.

stephenfuqua (approved):

stephenfuqua:
@lalrg can you do a final review, test, and the merge?

bradbanister:
> I've done this, adding a note to the readme.

Thanks!

src/lms-ds-loader/edfi_lms_ds_loader/migrator.py
```diff
@@ -19,18 +19,18 @@
...
+    "0003_create_user_tables",
+    "0004_create_section_tables",
+    "0005_create_assignment_tables",
+    "0006_create_section_association_tables",
+    "0007_create_assignment_submission_tables",
+    "0008_create_section_activity_tables",
+    "0009_create_system_activity_tables",
+    "0010_create_attendance_tables",
+    "0011_remove_startdate_enddate_from_sectionassociation",
+    "0012_add_mapping_columns_for_edfi_student_and_section",
```
lalrg, line 33: We could do it by convention, simply look for the .SQL files in the folder, split by '_' and get the first part, then sort and execute it, so that we don't need to hardcode the migration names. Not saying it should be done on this ticket, it is just an idea.

lalrg (approved):
Looks good to me

src/lms-ds-loader/edfi_lms_ds_loader/migrator.py
bradbanister: FYI, this is how the code in my end to end integration tests loads these files.

PR #215, opened 2021-07-02
[LMS-335] Update gc user harmonizer to use lms.EmailAddress

bradbanister:


stephenfuqua (approved):
Looks good and test plan is passing.

PR #216, opened 2021-07-06
[LMS-340] Derive late submission status from "late" flag

bradbanister:


src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignments.py
```diff
@@ -68,13 +68,10 @@ def coursework_to_assignments_dfs(
                 "dueTime.minutes",
             ]
         ]
-        filled_df: Optional[DataFrame] = due_date_df.fillna("0")
```
bradbanister, line 71: This was crashing on various oddities in the due date data from the live API during testing.

extension/Descriptors/SubmissionStatusDescriptor.xml
```diff
@@ -78,4 +78,16 @@
 		<Description>Student chose to "unsubmit" the assignment.</Description>
 		<Namespace>uri://ed-fi.org/edfilms/SubmissionStatusDescriptor/Google</Namespace>
 	</SubmissionStatusDescriptor>
+    <SubmissionStatusDescriptor>
+		<CodeValue>LATE</CodeValue>
```
lalrg, line 82: Is it necessary to put it in uppercase for the CodeValue column? I think it is easier to read, join and query when the values match in the different columns, so when possible I like to have them written in the same way,

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
```diff
@@ -4,10 +4,63 @@
...
-from pandas import DataFrame
+from pandas import DataFrame, isna
 from edfi_google_classroom_extractor.mapping.constants import SOURCE_SYSTEM
 
+# States returned by API
+CREATED_STATE = "CREATED"
+NEW_STATE = "NEW"
+RECLAIMED_STATE = "RECLAIMED_BY_STUDENT"
 TURNED_IN_STATE = "TURNED_IN"
+RETURNED_STATE = "RETURNED"
```
lalrg, line 15:
```suggestion
class API_STATES:
  CREATED_STATE = "CREATED"
  NEW_STATE = "NEW"
  RECLAIMED_STATE = "RECLAIMED_BY_STUDENT"
  TURNED_IN_STATE = "TURNED_IN"
  RETURNED_STATE = "RETURNED"
```
This can improve readability

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
```diff
@@ -4,10 +4,63 @@
...
+
+# States derived from API "late" flag
+LATE_STATE = "LATE"
+MISSING_STATE = "MISSING"
```
lalrg, line 19:
```suggestion
class LMS_STATES:
    LATE_STATE = "LATE"
    MISSING_STATE = "MISSING"
```

src/google-classroom-extractor/tests/mapping/assignment_submissions/test_derived_status.py
```diff
@@ -0,0 +1,283 @@
...
+
+        submissions_df = DataFrame.from_dict(
+            [
+                merged_dict(submission, BOILERPLATE),
+            ]
+        )
+        # act
+        return submissions_to_assignment_submissions_dfs(submissions_df)
+
+    def it_should_be_correct(assignment_submission_dict):
```
lalrg, line 74:
```suggestion
    def it_should_be_missing_state(assignment_submission_dict):
```

src/google-classroom-extractor/tests/mapping/assignment_submissions/test_derived_status.py
```diff
@@ -0,0 +1,283 @@
...
```
lalrg, line 97:
```suggestion
    def it_should_be_created_state(assignment_submission_dict):
```

src/google-classroom-extractor/tests/mapping/assignment_submissions/test_derived_status.py
```diff
@@ -0,0 +1,283 @@
...
```
lalrg, line 120:
```suggestion
    def it_should_be_missing_state(assignment_submission_dict):
```

src/google-classroom-extractor/tests/mapping/assignment_submissions/test_derived_status.py
```diff
@@ -0,0 +1,283 @@
...
```
lalrg, line 143:
```suggestion
    def it_should_be_created_state(assignment_submission_dict):
```

lalrg: I think it is easier to read when a test fails

src/google-classroom-extractor/tests/mapping/assignment_submissions/test_derived_status.py
```diff
@@ -0,0 +1,283 @@
...
```
lalrg, line 165:
```suggestion
    def it_should_be_created_state(assignment_submission_dict):
```

lalrg: same for the rest of the tests 😸

lalrg (changes requested):
I left some suggestions as comments.

extension/Descriptors/SubmissionStatusDescriptor.xml
stephenfuqua: The original code values were copied directly from the source systems, preserving the case in from the source system documentation. In this situation we are making up the values and I have no preference. We do assume that SQL Server has been setup with the default case insensitive collation.

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
```diff
@@ -4,10 +4,63 @@
...
+
+
+def derive_state(submission_row: Any) -> str:
```
stephenfuqua, line 22: Looks like `submission_row` is at least a `dict` rather than any `Any`...

stephenfuqua: No, I see now that this is not from `iterrows` but from `apply`. So the type should be `pandas.core.series.Series`, I think.

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignments.py
```diff
@@ -68,13 +68,10 @@ def coursework_to_assignments_dfs(
...
-        if filled_df is None:
-            raise ValueError  # fillna will never return None in this usage
-        coursework_df["DueDateTime"] = filled_df.apply(
-            lambda date_element: datetime(
-                *to_numeric(date_element, downcast="integer")
-            ),
+        coursework_df["DueDateTime"] = due_date_df.apply(
+            lambda date_element: ""
+            if date_element.isnull().any()
```
stephenfuqua, line 73: How quickly we forget things! For example, I forgot this lovely Python syntax of "a if (condition) else b" 😄 .

extension/Descriptors/SubmissionStatusDescriptor.xml
bradbanister: Yeah, I'm following the upper case nature of the statuses that come from the API. If we want to do something different, we should probably have a bigger discussion around normalization of incoming values.

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
bradbanister: Not sure it does. I'd think "API_STATES.NEW_STATE" would have redundant naming, so I'd want to drop the "_STATE" suffix, making it "API_STATES.NEW". Not much different, but longer than "NEW_STATE" without a readablility improvement. And on other modules I might want to import "from API_STATES import NEW", and now I'm using "NEW" where "NEW_STATE" would be clearer. Could use the shorter class name "STATE", but same issue with above import, while a plain "import STATE" results in "STATE.NEW". Not really a readability difference between "NEW_STATE" and "STATE.NEW".

src/google-classroom-extractor/tests/mapping/assignment_submissions/test_derived_status.py
bradbanister: 👍 for these

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignments.py
bradbanister: 😎

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
bradbanister: Nice catch, Any was a placeholder. Should be pandas Series.

lalrg (approved):
The code looks fine and it works as expected.

PR #217, opened 2021-07-07
[LMS-323] Integration Testing for Analytics Middle Tier with Extension Model

stephenfuqua:


utils/amt-integration-tests/.vscode/settings.json
```diff
@@ -0,0 +1,110 @@
+{
+    "python.analysis.extraPaths": ["./change_to_module_name", "./tests"],
```
lalrg, line 2: We should change this `"./change_to_module_name"`

utils/amt-integration-tests/pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+black = "^21.6b0"
+flake8 = "^3.9.2"
+pandas = "^1.3.0"
+SQLAlchemy = "1.4.20"
+pyodbc = "^4.0.31"
+pytest-bdd = "^4.1.0"
+lxml = "^4.6.3"
+
+[tool.poetry.dev-dependencies]
+
```
lalrg, line 20:
```suggestion
[tool.poetry.dependencies]
python = "^3.9"
pytest = "^6.2.4"
pandas = "^1.3.0"
SQLAlchemy = "1.4.20"
pyodbc = "^4.0.31"
pytest-bdd = "^4.1.0"
lxml = "^4.6.3"

[tool.poetry.dev-dependencies]
black = "^21.6b0"
flake8 = "^3.9.2"
```
I proposed only moving some of the packages, but if this will always be a development tool, almost all dependencies should be dev dependencies

utils/amt-integration-tests/.vscode/settings.json
stephenfuqua: Done

utils/amt-integration-tests/tests/data_helpers.py
```diff
@@ -0,0 +1,163 @@
...
+
+APPEND_OPTIONS = {"if_exists": "append", "index": False}
+
+
+def load_school(engine: engine.base.Engine, id: str) -> None:
+    ed_org = pd.DataFrame([{"EducationOrganizationId": id, "NameOfInstitution": id}])
+    ed_org.to_sql("EducationOrganization", **_get_edfi_options(engine))
+
+    school = pd.DataFrame([{"schoolid": id}])
+    school.to_sql("School", **_get_edfi_options(engine))
```
lalrg, line 76: Why are you using a dataframe instead of an engine.execute("sql script") or something like that?

utils/amt-integration-tests/pyproject.toml
stephenfuqua: Since it is _only_ a development tool, I didn't see any value in differentiating between non-dev and dev dependencies.

utils/amt-integration-tests/tests/data_helpers.py
stephenfuqua: Because this is easier than the alternative of writing a SQL statement by hand. Just look at how little code was needed to get this working ;-). Oh, and it will also work on either PostgreSQL or SQL Server, whereas if it was hand written it might only work on SQL Server.

utils/amt-integration-tests/tests/features/assignmentdim_view.feature
```diff
@@ -0,0 +1,93 @@
...
+            | SchoolKey        | 53                                   |
+            | SourceSystem     | Schoology                            |
+            | Title            | A discussion for 1053                |
+            | Description      | The description                      |
+            | StartDateKey     | 1053-07-08                           |
+            | EndDateKey       | 1053-07-09                           |
+            | DueDateKey       | 1053-07-09                           |
+            | MaxPoints        | 99                                   |
+            | SectionKey       | 53-abc-1053-1053-si-1503-Summer-1053 |
+            | GradingPeriodKey | r"[0-9]+-53-1053-07-01"              |
```
stephenfuqua, line 74: The first part of the `GradingPeriodKey` is a `GradingPeriodDescriptorId`. Of course we have no way of knowing what that ID will be in advance, so here I used `r"expression"` to signal that a regular expression is in use, and the first component `[0-9]+` signals that we are just expecting a number there. At this point, we're just trusting that the _correct_ number goes into this field. Doing anything else would require substantial effort and probably wouldn't be worth it.

utils/amt-integration-tests/tests/features/assignmentdim_view.feature
```diff
@@ -0,0 +1,93 @@
...
+            | LastModifiedDate | r"\d{4}-\d{2}-\d{2}"                 |
```
stephenfuqua, line 75: Another example, we're not inserting the LastModifiedDate. But now that I think about it, we _should_. Just because there is an default value on the table doesn't mean we have to use that default. In the requirements, this value should be coming from the `lmsx.Assignment` table. So we should go ahead and hard-code a value in that table insert so that we can confirm the right value comes out here.

stephenfuqua:
We're abandoning this branch.

PR #220, opened 2021-07-16
[LMS-331] sample LMS data for assignments & submissions

markramonDL:
sample LMS data to support importing of assignments and submissions.  assignments.csv, users.csv, sections.csv, section-assoc.csv, submissions.csv


stephenfuqua:
We need to re-organize the files so that they fit into the file system properly, including splitting into separate section files.

bradbanister:
Moving to a branch in the main repo

PR #226, opened 2021-07-27
[LMS-360] SubmissionDateTime is nullable

stephenfuqua:
To test this, simply copy the `extension/EdFi.Ods.Extensions.LMSX` files to `Ed-Fi-ODS-Implementation/Applications/EdFi.Ods.Extensions.LMSX`, add a reference in the Web API project to the LMSX project, and run initdev. Look at the PopulatedTemplate database - is the column nullable in the {{lmsx.AssignmentSubmission}} table?

Then try uploading the new `docs/starter-kit-sample` files into the populated template db, using the LMS-DS-Loader. Before the change that would have failed, with an error about inserting null into a non-nullable field. Now it is accepted.


lalrg (approved):
The code looks good and it is working as expected.

PR #227, opened 2021-07-27
[LMS-363] Insert SchoolId in lmsx.Assignment table

lalrg:


extension/EdFi.Ods.Extensions.LMSX/LMS-Harmonizer/1080-lms-assignment.sql
```diff
@@ -23,6 +23,7 @@ BEGIN
 		edfiSection.LocalCourseCode [LocalCourseCode],
 		edfiSection.[SessionName] [SessionName],
 		edfiSection.[SchoolYear] [SchoolYear],
+		edfiSection.SchoolId,
```
stephenfuqua, line 26: Is there anything you can update in an integration test?

stephenfuqua: We missed an opportunity to test that the fields are being mapped properly. But we can easily correct that: in `test_lmsassignments.py`, at line 106, we can test the individual field assignments, to make sure that each value in `lmsx.Assignment` was assigned correctly.

stephenfuqua (changes requested):
We need to improve the test coverage. If we had tested each field before then we might have noticed the error during the original development.

src/lms-harmonizer/tests_integration_sql/procs/mssql/harmonize_assignments/test_lmsassignments.py
```diff
@@ -98,11 +98,49 @@ def it_should_run_successfully(test_db_config: ServerConfig, source_system: str)
...
+        assert (
+            LMSAssignment["SessionName"] is not None
+        ), "It should map the SessionName"
+
+        assert (
+            LMSAssignment["SchoolYear"] is not None
+        ), "It should map the SchoolYear"
+
+        assert (
+            LMSAssignment["SchoolId"] is not None
```
stephenfuqua, line 138: The `SchoolId` value is 1. See line 291 in `mssql_loader.py`. Other values can be known based on constants. So we should only do the  "is not None" test for those values that are hard for us to know - that aren't hard-coded somewhere in our Python modules..

stephenfuqua: So please re-arrange so that the values that _can't_ easily be known - descriptor IDs are really the only ones - are at the end of the list. All of the other asserts should be above the comment, and should look for the specific known values.

src/lms-harmonizer/tests_integration_sql/procs/mssql/harmonize_assignments/test_lmsassignments.py
```diff
@@ -98,11 +98,48 @@ def it_should_run_successfully(test_db_config: ServerConfig, source_system: str)
...
+        assert (
+            int(LMSAssignment["SchoolYear"]) == 1
+        ), "It should map the SchoolYear"
+
+        assert (
+            int(LMSAssignment["SchoolId"]) == 1
+        ), "It should map the SchoolId"
+
+        assert (
+            LMSAssignment["Namespace"] is not None
```
stephenfuqua, line 141: In fact, we do know the source system - it was defined in a function call above.

src/lms-harmonizer/tests_integration_sql/procs/mssql/harmonize_assignments/test_lmsassignments.py
```diff
@@ -98,11 +98,48 @@ def it_should_run_successfully(test_db_config: ServerConfig, source_system: str)
...
+        assert (
+            LMSAssignment["SessionName"] == 'session name test'
+        ), "It should map the SessionName"
+
+        assert (
+            int(LMSAssignment["SchoolYear"]) == 1
+        ), "It should map the SchoolYear"
+
+        assert (
+            int(LMSAssignment["SchoolId"]) == 1
```
stephenfuqua, line 137: I am going to introduce some constants

src/lms-harmonizer/tests_integration_sql/procs/mssql/harmonize_assignments/test_lmsassignments.py
```diff
@@ -98,11 +98,48 @@ def it_should_run_successfully(test_db_config: ServerConfig, source_system: str)
...
-                int(LMSAssignment[0]["AssignmentIdentifier"]) == 1
-            )  # It is using the identity field from lms.Assignment
+            result = query(connection, "SELECT * from [lmsx].[Assignment]")
+
+        assert len(result) == 1, "There should be one result."
+
+        LMSAssignment = result[0]
+
+        assert (
+            int(LMSAssignment["AssignmentIdentifier"]) == 1
```
stephenfuqua, line 108: This is not correct. However, it _is_ what the stored procedure is doing - so here's another bug! This is admittedly confusing; we did a poor job of naming fields. The data type gives it away though:

image

image

If the Source System were to make a POST request to the API, it would be inserting its _source system identifier_ as the `AssignmentIdentifier`.

stephenfuqua (approved):

extension/EdFi.Ods.Extensions.LMSX/LMS-Harmonizer/1080-lms-assignment.sql
```diff
@@ -7,26 +7,25 @@ CREATE OR ALTER PROCEDURE [lms].[harmonize_assignment] @SourceSystem nvarchar(25
...
-		lmsAssignment.[AssignmentDescription] [AssignmentDescription],
-		lmsAssignment.[StartDateTime] [StartDateTime],
-		lmsAssignment.[EndDateTime] [EndDateTime],
-		lmsAssignment.[DueDateTime] [DueDateTime],
-		lmsAssignment.[MaxPoints] [MaxPoints],
-		edfiSection.SectionIdentifier [SectionIdentifier],
-		edfiSection.LocalCourseCode [LocalCourseCode],
-		edfiSection.[SessionName] [SessionName],
-		edfiSection.[SchoolYear] [SchoolYear],
+		lmsAssignment.SourceSystemIdentifier as AssignmentIdentifier,
```
stephenfuqua, line 12: I like to have `as` - although it is unnecessary, I find it easier to read: the syntax highlighting helps me see that an alias is being created.

Avoiding brackets whenever possible makes it easier to port the code to PostgreSQL in the future. And it is easier to write without brackets.

extension/EdFi.Ods.Extensions.LMSX/LMS-Harmonizer/1080-lms-assignment.sql
```diff
@@ -39,11 +38,11 @@ BEGIN
 			ON sourceSystemDescriptor.CodeValue = lmsAssignment.SourceSystem
 
 		INNER JOIN lmsx.LMSSourceSystemDescriptor
-		ON sourceSystemDescriptor.DescriptorId  = LMSSourceSystemDescriptor.LMSSourceSystemDescriptorId
+		    ON sourceSystemDescriptor.DescriptorId  = LMSSourceSystemDescriptor.LMSSourceSystemDescriptorId
 
 		INNER JOIN edfi.Descriptor assignmentCatDescriptor
 			ON assignmentCatDescriptor.CodeValue = lmsAssignment.AssignmentCategory
-			AND assignmentCatDescriptor.Namespace = 'uri://ed-fi.org/edfilms/AssignmentCategoryDescriptor/' + lmsAssignment.SourceSystem
+			AND assignmentCatDescriptor.Namespace = 'uri://ed-fi.org/edfilms/AssignmentCategoryDescriptor/' + @SourceSystem
```
stephenfuqua, line 45: I was worried that this might cause problems for indexing. Building the string from a constant, instead of from a value in another table, makes this cleaner anyway.

extension/EdFi.Ods.Extensions.LMSX/LMS-Harmonizer/1080-lms-assignment.sql
```diff
@@ -66,7 +65,7 @@ BEGIN
 		,[SchoolYear]
 		,[SchoolId]
 		,[Namespace])
-	select
+	SELECT
```
stephenfuqua, line 68: All caps isn't necessary. However, the use of lowercase in just a few places was inconsistent with the rest of the file. So I updated it.

extension/EdFi.Ods.Extensions.LMSX/LMS-Harmonizer/1080-lms-assignment.sql
stephenfuqua: Oh, and this line 12 is the real reason that I was modifying the file. Note that it is now pulling from `SourceSystemIdentifier` instead of `AssignmentIdentifeir`.

extension/EdFi.Ods.Extensions.LMSX/LMS-Harmonizer/1090-lms-assignment-submissions.sql
```diff
@@ -20,9 +20,16 @@ BEGIN
 		lmsSubmission.LastModifiedDate,
 		lmsSubmission.DeletedAt
 	INTO #ALL_SUBMISSIONS
-	FROM LMS.AssignmentSubmission lmsSubmission
-	INNER JOIN LMSX.Assignment lmsxAssignment
-		ON lmsSubmission.AssignmentIdentifier = lmsxAssignment.AssignmentIdentifier
```
stephenfuqua, line 25: We definitely made a mistake in our column naming - this _looks_ correct but is in fact wrong. the value on the left is an integer, while the value on the right is a string - it is the Assignment's Source System Identifier. In order to write the correct query, I had to add a join from `lms.AssignmentSubmission` to `lms.Assignment`, and then the join to `lmsx.Assignment` (the extension table) goes through `lms.Assignment` instead of trying to directly join `lms.AssignmentSubmission` to `lmsx.Assignment`.

extension/EdFi.Ods.Extensions.LMSX/LMS-Harmonizer/1080-lms-assignment.sql
lalrg: yes it is better, I wrote the harmonize for the assignments and cannot remember why I used the assignment identifier instead of the source system identifier

stephenfuqua: Easy mistake to make, as we mixed up the column names: in one place `AssignmentIdentifier` is an auto-increment int, and in another place it is a varchar. But that's not obvious just from the column name.

PR #228, opened 2021-08-03
[LMS-359] Support for encrypted connections in the SQL Adapter

stephenfuqua:
This is the first of at least two pull requests for this ticket


lalrg (approved):
Code looks fine and tests coverage is fine too.

PR #230, opened 2021-08-04
[LMS-359] Support encrypted connections in the harmonizer and ds-loader

stephenfuqua:
@lalrg  Suggestions for testing the LMS DS Loader:

1. Setup the DS loader to upload the large Schoology files.
2. In SSMS, connect to your destination database and prepare to run the following query:

```sql
   SELECT   
    c.session_id, c.net_transport, c.encrypt_option,   
    c.auth_scheme, s.host_name, s.program_name,   
    s.client_interface_name, s.login_name, s.nt_domain,   
    s.nt_user_name, s.original_login_name, c.connect_time,   
    s.login_time   
FROM sys.dm_exec_connections AS c  
JOIN sys.dm_exec_sessions AS s  
    ON c.session_id = s.session_id  
```

3. Enable encryption without trusting server certificates. You should see an error message, because the localhost certificate is self-signed and therefore untrusted. This means the `--encrypt` flag is working.
4. Now trust the server certificate and re-run. This should take a little bit of time because of the large LMSUser file.
5. Quickly switch back to SSMS and run the query above. Confirm that the `ENCRYPT_OPTION` column reports `TRUE`.

You can run the same tests for the LMS Harmonizer, but if it runs very quickly then it may be hard to see the open connection with that SQL query above.


lalrg (approved):
Everything looks and works just fine! Since I made the last commit with an update in the readme, I will approve but leave the merge to @stephenfuqua

PR #235, opened 2021-08-13
[LMS-371] Get courses directly from /api/v1/courses

lalrg:
Remove the dependency on accounts so the course does not get duplicated by having a subaccount set.

I already created the subaccount in the test environment, you can check the steps to reproduce and look at the development subaccount, it has two courses referenced. So, simply run the code and it should be working without any error nor duplicated data.


lalrg:
This approach won't work since with it, we don't have access to all the courses that we need. the ticket will be addressed on #237

PR #236, opened 2021-08-15
Warning when no courses found

stephenfuqua:


stephenfuqua:
Now that I have found the root cause of why there were no courses, and determined it is not a permissions problem, I'm going to close this PR.

PR #237, opened 2021-08-15
[LMS-371] Remove duplicate objects before sync

stephenfuqua:


lalrg (approved):
The bug described is solved, and the code looks fine.

PR #239, opened 2021-08-19
[LMS-355] Add validations to avoid duplicated rows and add integration tests

lalrg:


src/lms-harmonizer/edfi_lms_harmonizer/scripts/mssql/0014_view_assignment_submissions_exceptions.sql
```diff
@@ -7,5 +7,13 @@ CREATE VIEW lmsx.assignment_submissions_exceptions AS
     SELECT * FROM LMS.AssignmentSubmission lmsAssignmentSubmission
     WHERE NOT EXISTS(
             SELECT 1 FROM lmsx.AssignmentSubmission lmsxassignmentsubmission
-                where lmsxassignmentsubmission.AssignmentSubmissionIdentifier = lmsAssignmentSubmission.SourceSystemIdentifier
+                INNER JOIN lmsx.Assignment lmsxassginment
```
stephenfuqua, line 10: In general, I find it a little easier to read like this:

```INNER JOIN lmsx.Assignment as lmsxassignment```

the `as` keyword in SSMS makes it easier to quickly scan the query and find the aliases.

No need to spend time changing this right now. Just something to think about for future work.

stephenfuqua (approved):

PR #241, opened 2021-08-30
[LMS-390] Add the upcoming status for submissions when needed in Canvas

lalrg:


src/canvas-extractor/edfi_canvas_extractor/client_facade.py
```diff
@@ -226,7 +226,8 @@ def extract_enrollments(
         local_enrollments: List[Enrollment] = list(
             enrollmentsApi.request_enrollments_for_section(section)
         )
-        if len(list(local_enrollments)) < 1:
+        filtered_enrollments = [enrollment for enrollment in local_enrollments if enrollment.enrollment_state == "active"]
+        if len(list(filtered_enrollments)) < 1:
             logger.info(
                 "Skipping enrollments for section id %s - No data returned by API",
```
stephenfuqua, line 232: Is this comment meaningful? Maybe useful to have _something_, but I don't think this message is very good anymore, especially in light of the adjustment made on line 229.

We still write out a Section Association file, so we're not really "skipping enrollments".

```suggestion
                "There are no no active section associations for section id %s.",
```

src/canvas-extractor/edfi_canvas_extractor/mapping/assignments.py
```diff
@@ -122,6 +122,8 @@ def map_to_udm_assignments(
         "LMSSectionSourceSystemIdentifier"
     ].astype("string")
 
+    assignments_df = assignments_df.drop_duplicates()
```
stephenfuqua, line 125: Is this actually necessary?

lalrg: Yes... If you remove it, it duplicates records

src/canvas-extractor/edfi_canvas_extractor/client_facade.py
```diff
@@ -198,7 +198,7 @@ def extract_submissions(
         for assignment_id, submissions_df in submissions_dfs_by_assignment_id.items():
             section_id = str(section.id)
             submissions_df = submissionsMap.map_to_udm_submissions(submissions_df, section_id)
-            export[(section_id, str(assignment_id))] = submissions_df
+            export[(section_id, f"{section_id}-{str(assignment_id)}")] = submissions_df
```
stephenfuqua, line 201: This causes the output directory to be like `assignment=104-111` instead of `assignment=111`. That's a good change, since the actual `SourceSystemId` is the two part value. Is it possible to get that value from the DataFrame instead of having to build it here?

src/canvas-extractor/edfi_canvas_extractor/client_facade.py
```diff
@@ -226,9 +226,10 @@ def extract_enrollments(
         local_enrollments: List[Enrollment] = list(
             enrollmentsApi.request_enrollments_for_section(section)
         )
-        if len(list(local_enrollments)) < 1:
+        filtered_enrollments = [enrollment for enrollment in local_enrollments if enrollment.enrollment_state == "active"]
+        if len(list(filtered_enrollments)) < 1:
             logger.info(
-                "Skipping enrollments for section id %s - No data returned by API",
+                "There are no no active section associations for section id %s.",
```
stephenfuqua, line 232: Oops!

```suggestion
                "There are no active section associations for section id %s.",
```

src/canvas-extractor/edfi_canvas_extractor/mapping/submissions.py
```diff
@@ -19,12 +19,14 @@ def _get_date_formated(date) -> str:
...
+    if row["late"] is True:
         return "late"
-    if row["missing"] == 1:
+    if row["missing"] is True:
         return "missing"
-    if row["graded_at"] is not None:
+    if pd.isnull(row["graded_at"]) is False:
         return "graded"
+    if pd.isnull(row["SubmissionDateTime"]) is False:
+        return "upcoming"
```
stephenfuqua, line 29: I created a new assignment in section 103 (ENG-1), with due date set to 2021-12-17. The output file has "on-time" for all of the records instead of "upcoming". Here's an example raw response from the GET request:

```javascript
{
        "assignment_id": 1221,
        "id": 11464,
        "grading_period_id": null,
        "body": null,
        "url": null,
        "grade": null,
        "score": null,
        "submitted_at": null,
        "user_id": 111,
        "submission_type": null,
        "workflow_state": "unsubmitted",
        "grade_matches_current_submission": true,
        "graded_at": null,
        "grader_id": null,
        "attempt": null,
        "cached_due_date": "2021-12-18T05:59:59Z",
        "excused": null,
        "late_policy_status": null,
        "points_deducted": null,
        "extra_attempts": null,
        "posted_at": null,
        "late": false,
        "missing": false,
        "seconds_late": 0,
        "entered_grade": null,
        "entered_score": null,
        "preview_url": "https://edfialliance.instructure.com/courses/104/assignments/1221/submissions/111?preview=1&version=0",
        "anonymous_id": "Lwc1w"
    },
```

In the same section, assignment 114 has some records like this one, and they are showing up with status "upcoming" instead of "missing". So I wonder if the change on line 24 was incorrect.

```javascript
    {
        "assignment_id": 114,
        "id": 11456,
        "grading_period_id": null,
        "body": null,
        "url": null,
        "grade": null,
        "score": null,
        "submitted_at": null,
        "user_id": 108,
        "submission_type": null,
        "workflow_state": "unsubmitted",
        "grade_matches_current_submission": true,
        "graded_at": null,
        "grader_id": null,
        "attempt": null,
        "cached_due_date": "2021-05-01T06:01:00Z",
        "excused": null,
        "late_policy_status": null,
        "points_deducted": null,
        "extra_attempts": null,
        "posted_at": null,
        "late": false,
        "missing": true,
        "seconds_late": 10510801,
        "entered_grade": null,
        "entered_score": null,
        "preview_url": "https://edfialliance.instructure.com/courses/104/assignments/114/submissions/108?preview=1&version=0",
        "anonymous_id": "yxUaK"
    }
```

Finally, I created my own Submission on the assignment above. The API response has `"late": true`, but the status in the extracted file is "upcoming" instead of "late."

Finally, shouldn't line 28 be looking for cases where the `SubmissionDateTime` _is null_ instead of _not is null_?

stephenfuqua (changes requested):

src/canvas-extractor/edfi_canvas_extractor/client_facade.py
lalrg: The issue with that approach is that here we are in the extraction part, not the mapping or the writing part, so we don't have access to the mapped SourceSystemIdentifier, and we need to build it here.

src/canvas-extractor/edfi_canvas_extractor/client_facade.py
lalrg: 😆 looks like if the log were singing

src/canvas-extractor/edfi_canvas_extractor/client_facade.py
stephenfuqua: Maybe we should move that code from the mapper here then, so that we can take this action only once. Something like

```python
assignment_source_system_identifier = f"{section_id}-{str(assignment_id)}"
submissions_df["assignment_id"] = assignment_source_system_identifier
```

Side note: in other cases, didn't we use a standard like `f"{section_id}#{str(assignment_id)}"` instead of `f"{section_id}-{str(assignment_id)}"`?

lalrg: I think so, maybe for section activity or something like that

lalrg: changes applied

src/canvas-extractor/edfi_canvas_extractor/client_facade.py
```diff
@@ -197,8 +197,10 @@ def extract_submissions(
 
         for assignment_id, submissions_df in submissions_dfs_by_assignment_id.items():
             section_id = str(section.id)
+            assignment_source_system_identifier = f"{section_id}-{str(assignment_id)}"
+            submissions_df["assignment_id"] = assignment_source_system_identifier
             submissions_df = submissionsMap.map_to_udm_submissions(submissions_df, section_id)
-            export[(section_id, str(assignment_id))] = submissions_df
+            export[(section_id, f"{section_id}-{str(assignment_id)}")] = submissions_df
```
stephenfuqua, line 203:
```suggestion
            export[(section_id, f"{assignment_source_system_identifier}")] = submissions_df
```

src/canvas-extractor/edfi_canvas_extractor/mapping/submissions.py
```diff
@@ -19,12 +19,14 @@ def _get_date_formated(date) -> str:
 
 
 def _get_status(row: pd.Series):
-    if row["late"] == 1:
+    if row["late"] is True:
         return "late"
```
stephenfuqua, line 23: This still isn't working. Look at section 103, assignment 114:  `GET 'https://edfialliance.instructure.com/api/v1/sections/103/assignments/114/submissions'` . Submission 11456 was late and is reported as late by the API, but the extractor output says that it is on-time. And Submission 131 says "upcoming" when the status is actually "missing".

I think changing from `== 1` to `is True` was a mistake.

lalrg: even by reverting it, it keeps saying it is on time, I will look into it a little bit more

src/canvas-extractor/edfi_canvas_extractor/client_facade.py
lalrg: 🤦🏼‍♂️

src/canvas-extractor/edfi_canvas_extractor/mapping/submissions.py
lalrg: Fixed

src/canvas-extractor/edfi_canvas_extractor/mapping/submissions.py
```diff
@@ -19,12 +19,14 @@ def _get_date_formated(date) -> str:
...
 def _get_status(row: pd.Series):
-    if row["late"] == 1:
+    if row["late"] == 'True':
         return "late"
-    if row["missing"] == 1:
+    if row["missing"] == 'True':
         return "missing"
-    if row["graded_at"] is not None:
+    if pd.isnull(row["graded_at"]) is False:
         return "graded"
```
stephenfuqua, line 27: We no longer have "on-time". And that's ok - it looks like we had two different [descriptors](https://github.com/Ed-Fi-Alliance-OSS/LMS-Toolkit/blob/main/extension/Descriptors/SubmissionStatusDescriptor.xml) for the same thing: "graded" and "on-time". 

In Analytics Middle Tier, both status are correctly mapped to the "On time" descriptor constant, so that's good.

Can you go ahead and remove the "on-time" Descriptor from the XML file? Or remove "graded" and change above to "on-time"? I don't have a preference between the two.

stephenfuqua (approved):

PR #244, opened 2021-09-07
[LMS-406] Fix Google Extractor submission Datetime mapping

lalrg:


src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
```diff
@@ -63,6 +66,39 @@ def derive_state(submission_row: Series) -> str:
     return api_state
 
 
+def _get_submission_datetime(submission_row: Series) -> str:
+    if (
+        submission_row["state"] == TURNED_IN_STATE
+        or submission_row["state"] == RETURNED_STATE
```
stephenfuqua, line 72: "Returned" means the teacher has graded the assignment. This isn't the submission date/time - should only be looking at `TURNED_IN_STATE`.

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
```diff
@@ -3,10 +3,13 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
-from typing import Dict, Tuple, Any
+from typing import Dict, List, Tuple, Any
+import ast
```
stephenfuqua, line 7: What is `ast`?

lalrg: The library: https://docs.python.org/3/library/ast.html
The function: https://docs.python.org/3/library/ast.html#ast.literal_eval

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
stephenfuqua: What do you think about something like this?

If state == TURNED_IN { return updateTime }
if state == RETURNED { look in history and find the newest TURNED_IN entry }
else { it has not been turned in }

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
```diff
@@ -63,6 +66,39 @@ def derive_state(submission_row: Series) -> str:
...
+        status_is_returned: bool = status_record["state"] == RETURNED_STATE
+        status_is_reclaimed_by_student: bool = status_record["state"] == RECLAIMED_STATE
+
+        if status_is_reclaimed_by_student:
+            return ""
+
+        if status_is_turned_in or status_is_returned:
+            return status_record["stateTimestamp"]
+
+    return ""
```
stephenfuqua, line 99: Some experimental code... not sure if this is any better, just different. I just needed some practice in Python!

```python
df = pd.DataFrame(
    [r["stateHistory"] for r in submission_row["submissionHistory"] if "stateHistory" in r]
)

df.sort_values(by=["stateTimestamp"], inplace=True,  ascending=False)

return df[df["state"] == "TURNED_IN"].iloc[0]["stateTimestamp"]
```

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
lalrg: ✅

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
stephenfuqua: I was looking on pypi and saw something really strange. This makes more sense now.

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
```diff
@@ -63,6 +66,35 @@ def derive_state(submission_row: Series) -> str:
...
+        for status_item in status_history
+        if "stateHistory" in status_item.keys()
+    ]
+    status_history = sorted(
+        status_history, key=lambda k: k["stateTimestamp"], reverse=True
+    )
+
+    for status_record in status_history:
+        status_is_turned_in: bool = status_record["state"] == TURNED_IN_STATE
+        status_is_reclaimed_by_student: bool = status_record["state"] == RECLAIMED_STATE
```
stephenfuqua, line 87: You could simplify this algorithm at the top: if the `submission_row["state"]` is not `TURNED_IN` or `RETURNED`, then just return None. That would allow us to skip several lines of code when the submission is in any other status.

lalrg: ✅

src/google-classroom-extractor/edfi_google_classroom_extractor/mapping/assignment_submissions.py
lalrg: I don't know if `If state == TURNED_IN { return updateTime }` is a good idea... What happens if a student turns in the submission and for some reason it gets updated after it without a change in the status? It is okay to only enter in the logic when the status is turned_in or returned, but we still may want to check the status history to have an accurate date

bradbanister (approved):
Compared to old data pulls, looks good.

PR #249, opened 2021-11-17
[LMS-233] Loader support for Postgres: Users

lalrg:
I decided to keep the integration tests with the mssql name, because we could want to create integration tests in the future.


src/lms-ds-loader/edfi_lms_ds_loader/Sql_lms_operations.py
```diff
@@ -62,7 +72,8 @@ def truncate_staging_table(self, table: str) -> None:
         assert table.strip() != "", "Argument `table` cannot be whitespace"
 
         # Note: for postgresql we'll want `TRUNCATE TABLE {staging} RESTART IDENTITY`
-        self._exec(f"TRUNCATE TABLE lms.stg_{table};")
+        if self.engine == DbEngine.MSSQL:
+            self._exec(MS_builder.truncate_stg_table(table))
```
stephenfuqua, line 76: This command should be the same on both SQL Server and PostgreSQL.

src/lms-ds-loader/edfi_lms_ds_loader/Sql_lms_operations.py
```diff
@@ -77,9 +88,8 @@ def disable_staging_natural_key_index(self, table: str) -> None:
 
         assert table.strip() != "", "Argument `table` cannot be whitespace"
 
-        self._exec(
-            f"ALTER INDEX IX_stg_{table}_Natural_Key on lms.stg_{table} DISABLE;"
-        )
+        if self.engine == DbEngine.MSSQL:
+            self._exec(MS_builder.disable_staging_natural_key_index(table))
```
stephenfuqua, line 92: Looks like PostgreSQL doesn't have clean functionality for achieving this. Probably need to just drop and recreate the index.

src/lms-ds-loader/edfi_lms_ds_loader/Sql_lms_operations.py
```diff
@@ -139,26 +148,9 @@ def insert_new_records_to_production(self, table: str, columns: List[str]) -> No
...
-            lms.{table}
-        WHERE
-            SourceSystemIdentifier = stg.SourceSystemIdentifier
-        AND
-            SourceSystem = stg.SourceSystem
-    )
-"""
+        statement = ""
+        if self.engine == DbEngine.MSSQL:
+            statement = MS_builder.insert_new_records_to_production(table, column_string)
```
stephenfuqua, line 153: I think the SQL statement above will work on both systems, or could be rewritten to work on both.

src/lms-ds-loader/edfi_lms_ds_loader/Sql_lms_operations.py
```diff
@@ -188,33 +180,10 @@ def insert_new_records_to_production_for_user_relation(
...
-  WHERE
-    SourceSystemIdentifier = stg.SourceSystemIdentifier
-  AND
-    SourceSystem = stg.SourceSystem
-)
-"""
+        statement = ""
+        if self.engine == DbEngine.MSSQL:
+            statement = MS_builder.insert_new_records_to_production_for_user_relation(
+                table, insert_columns, select_columns)
```
stephenfuqua, line 186: Again, I think this would work on both or could be rewritten to work on both.

stephenfuqua: After reading the entire PR, I'm fine with having two separate files even if when some of the SQL could be reused.

src/lms-ds-loader/edfi_lms_ds_loader/Sql_lms_operations.py
stephenfuqua: https://dba.stackexchange.com/a/116805/183504

Might want to read all of the index definitions, hold them in memory. Drop them. In the "enable" step, just re-execute the create statement. That way the SQL statement that is currently in the middle of a script file doesn't need to be repeated in Python.

src/lms-ds-loader/edfi_lms_ds_loader/Sql_lms_operations.py
lalrg: I guess some of them work for both, and yes, we could call the methods of the MS_builder when engine == MSSQL or POSTGRES and the SQL applies, the main idea that I had, was to make it somehow extensible in case we needed new statements for some other engine in the future.

src/lms-ds-loader/edfi_lms_ds_loader/db_operations/MSSQL_sql_builder.py
```diff
@@ -187,38 +187,45 @@ def insert_new_records_to_production_for_assignment_and_user_relation(
 
 
 def insert_new_records_to_production_for_attendance_events(
-        table: str,
         insert_columns: str,
         select_columns: str) -> str:
     return f"""
 INSERT INTO
-    lms.{table}
+    lms.LMSUserAttendanceEvent
```
stephenfuqua, line 194: This change seems fine... 

Do you remember why we made this a template before, instead of hard-coding the name as you have now done?

src/lms-ds-loader/edfi_lms_ds_loader/db_operations/MSSQL_sql_builder.py
```diff
@@ -187,38 +187,45 @@ def insert_new_records_to_production_for_assignment_and_user_relation(
...
 (
-    AssignmentIdentifier,
-    LMSUserIdentifier,{insert_columns}
+    LMSSectionIdentifier,
+    LMSUserIdentifier,
+    LMSUserLMSSectionAssociationIdentifier,{insert_columns}
```
stephenfuqua, line 198: We overlooked a column? And now you're fixing the bug?

src/lms-ds-loader/edfi_lms_ds_loader/db_operations/MSSQL_sql_builder.py
```diff
@@ -187,38 +187,45 @@ def insert_new_records_to_production_for_assignment_and_user_relation(
...
 def insert_new_records_to_production_for_attendance_events(
-        table: str,
         insert_columns: str,
         select_columns: str) -> str:
     return f"""
 INSERT INTO
-    lms.{table}
+    lms.LMSUserAttendanceEvent
 (
-    AssignmentIdentifier,
```
stephenfuqua: We previously inserted an `AssignmentIdentifier` into the attendance events table???

src/lms-ds-loader/tests/helpers/test_argparser.py
```diff
@@ -9,7 +9,7 @@
 
 from edfi_lms_ds_loader.helpers.argparser import parse_main_arguments, MainArguments
 from edfi_lms_ds_loader.helpers.constants import DbEngine, LOG_LEVELS
-from edfi_lms_ds_loader.mssql_lms_operations import MssqlLmsOperations
+from edfi_lms_ds_loader.Sql_lms_operations import SqlLmsOperations
```
stephenfuqua, line 12: Should this be lowercased?
```suggestion
from edfi_lms_ds_loader.sql_lms_operations import SqlLmsOperations
```

src/lms-ds-loader/edfi_lms_ds_loader/db_operations/MSSQL_sql_builder.py
lalrg: I was returning the wrong sql for this method and I fixed it

src/lms-ds-loader/tests/helpers/test_argparser.py
lalrg: yes

src/lms-ds-loader/edfi_lms_ds_loader/db_operations/MSSQL_sql_builder.py
lalrg: This also answers the previous comment on why I hardcoded the LMSUserAttendanceEvent, it does not have to be dynamic for this method and I am just fixing it, I realized of the error while fixing the tests

src/lms-ds-loader/edfi_lms_ds_loader/Sql_lms_operations.py
lalrg: There seem to be ways to disable indexes, but the general recommendation is to delete them and recreate them later.

src/lms-ds-loader/tests/helpers/test_argparser.py
lalrg: ✔️

stephenfuqua (approved):

PR #256, opened 2022-01-24
[LMS-455] Test for migrationjournal existence directly

bradbanister:


bradbanister:
Tested on both mssql and postgresql, from empty db, to partially migrated, to fully migrated.

stephenfuqua (approved):

PR #263, opened 2022-01-28
[LMS-436] Improved Integration Testing

stephenfuqua:


stephenfuqua:
When squashing this, we don't need to keep all the commit history with "Diagnosing..." etc.

src/lms-ds-loader/tests_integration_mssql/conftest.py
```diff
@@ -15,23 +19,85 @@
...
 )
 
+load_dotenv()
 
-def _new_mssql_adapter() -> Adapter:
-    # TODO: create method for injecting these values, for test
-    return create_mssql_adapter_with_integrated_security(
-        "localhost", "test_integration_lms_toolkit", 1433, False, False
+
+def pytest_addoption(parser):
```
stephenfuqua, line 28: The build.py script has command arguments for passing values into the tests, to support the Harmonizer. But the same scripts should also work with the DS Loader integration tests. This forced me to add the parser rules here, otherwise executing the DS Loader integration tests from build.py fails. So now I've actually taken care of https://tracker.ed-fi.org/browse/LMS-437

bradbanister (approved):
Looks good, all tests pass locally

PR #269, opened 2022-04-20
[LMS-479] Improve date handling

stephenfuqua:


src/canvas-extractor/edfi_canvas_extractor/mapping/authentication_events.py
```diff
@@ -76,10 +78,9 @@ def map_to_udm_system_activities(authentication_events: pd.DataFrame) -> pd.Data
     df["ActivityStatus"] = "active"
     df["ParentSourceSystemIdentifier"] = ""
     df["ActivityTimeInMinutes"] = ""
-    df["SourceCreateDate"] = df["ActivityDateTime"]
     df["SourceLastModifiedDate"] = ""
-    df["ActivityDateTime"] = pd.to_datetime(df["ActivityDateTime"]).dt.strftime(
-        constants.DATE_FORMAT
-    )
```
stephenfuqua, line 83: Previously, "SourceCreateDate" copied from "ActivityDateTime" before the conversion. Now the conversion occurs first, using a shared function.

src/canvas-extractor/tests/mapping/test_users.py
```diff
@@ -76,7 +76,7 @@ def test_then_email_address_is_mapped(result):
             assert result.at[0, "EmailAddress"] == "Kyle.Hughes@studentgps.org"
 
         def test_then_it_should_have_empty_SourceCreateDate(result):
-            assert result.at[0, "SourceCreateDate"] == "2020/09/14 11:54:18"
+            assert result.at[0, "SourceCreateDate"] == "2020-09-14 11:54:18"
```
stephenfuqua, line 79: Side effect of standardizing the date conversion: found an inconsistency in our previous handling.

stephenfuqua:
@andonyns @bradbanister please review. See https://tracker.ed-fi.org/browse/LMS-479 for context.

src/canvas-extractor/tests/mapping/test_users.py
bradbanister: Nice

bradbanister:
It looks like an invalid datetime will pass through as "NaT" to the CSV. Is that correct, and is that what we want? Do we define "NaT" as a valid value in CSV datatime fields?

docs/demonstration/README.md
```diff
@@ -3,8 +3,14 @@
 These notebooks demonstrate how to install and run the extractors and the
 Learning Management System Data Store Loader (LMS-DS Loader) in a Jupyter
 notebook. The same code can be copied over into a stand-alone Python script
-instead of running inside the notebook. Each notebook has its own
-getting-started instructions.
+instead of running inside the notebook.
+
+Before running the notebook:
+
+* Be sure that you have Python 3.9+ and Poetry
```
andonyns, line 10: In https://github.com/Ed-Fi-Alliance-OSS/LMS-Toolkit/blob/main/docs/demonstration/canvas-e2e.ipynb?short_path=68d8099#L13 and other notebooks, the instructions request for Python 3.9.x. I think these change suggests that you can use versions such as Python 3.10 with the LMS toolkit. Not sure if this is an actual requirement but seems like a mismatch.

andonyns (changes requested):
After running on main, it runs successfully. Then, after running in the current branch, I'm getting the following error:

File "C:\Users\anunez\pypoetry\Cache\virtualenvs\edfi-canvas-extractor-nRK1rodj-py3.9\lib\site-packages\pandas\core\indexes\accessors.py", line 509, in __new__
    raise AttributeError("Can only use .dt accessor with datetimelike values")
AttributeError: Can only use .dt accessor with datetimelike values
2022-04-21 17:10:44,943 - CRITICAL - edfi_canvas_extractor.extract_facade - Unable to continue file generation because the load of Students failed. Please review the log for more information.

I will provide the full log to research. This was executed with the latest changes after the fix from [Better handling for weird non-date strings](https://github.com/Ed-Fi-Alliance-OSS/LMS-Toolkit/pull/269/commits/c0c098b776f618a91a6bf5c00917614ecf509b69)

stephenfuqua:
> It looks like an invalid datetime will pass through as "NaT" to the CSV. Is that correct, and is that what we want? Do we define "NaT" as a valid value in CSV datetime fields?

Arg, no. I was in too much hurry and didn't look at the actual output. Unit testing showed it being converted to empty string. I'll need to revisit and figure out why.

docs/demonstration/README.md
stephenfuqua: Good call, we haven't tested with 3.10 so no idea if there is some kind of conflict there.

andonyns (approved):

bradbanister:
Agree, looks good.

PR #271, opened 2022-09-12
[LMS-468] Rewrite Canvas Extractor to use GraphQL for Sections

sgonzalezgap:
Its connecting to different files from __main__.py to start the process.
I didnt delete any of the previous files.

### DIFF
--- a/.github/workflows/test-canvas.yml
+++ b/.github/workflows/test-canvas.yml
@@ -34,6 +34,12 @@ jobs:
               with:
                   python-version: "3.9.x"
 
+            - name: Install Poetry
+              uses: Gr1N/setup-poetry@462ac83c852d49e282a1233c4c24c5411696e7c7
+              with:
+                  poetry-version: 1.1.7
+            - run: poetry --version
+
             - name: Get virtual env path
               id: venv-path
               run: |
--- a/.github/workflows/test-canvas.yml
+++ b/.github/workflows/test-canvas.yml
@@ -5,61 +5,56 @@
 
 name: Canvas Extractor - Test
 on:
-  pull_request:
-    paths:
-      - 'src/canvas-extractor/**'
-      - '!src/canvas-extractor/**/*.md'
-  push:
-    paths:
-      - 'src/canvas-extractor/**'
-      - '!src/canvas-extractor/**/*.md'
-    branches:
-      - main
-  workflow_dispatch:
+    pull_request:
+        paths:
+            - "src/canvas-extractor/**"
+            - "!src/canvas-extractor/**/*.md"
+    push:
+        paths:
+            - "src/canvas-extractor/**"
+            - "!src/canvas-extractor/**/*.md"
+        branches:
+            - main
+    workflow_dispatch:
 
 jobs:
-  test-canvas-extractor:
-    name: Run unit, style, and type checks
-    runs-on: ubuntu-20.04
-    env:
-        PACKAGE_NAME: canvas-extractor
-        SRC_DIR: src/canvas-extractor
-        LOCK_FILE: src/canvas-extractor/poetry.lock
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+    test-canvas-extractor:
+        name: Run unit, style, and type checks
+        runs-on: ubuntu-20.04
+        env:
+            PACKAGE_NAME: canvas-extractor
+            SRC_DIR: src/canvas-extractor
+            LOCK_FILE: src/canvas-extractor/poetry.lock
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
 
-      - name: Install Python 3.9
-        uses: actions/setup-python@3105fb18c05ddd93efea5f9e0bef7a03a6e9e7df
-        with:
-          # This one needs to be hard-coded to avoid 3.9.5, due to a bug
-          # in one of the dependencies
-          python-version: '3.9.4'
+            - name: Install Python 3.9
+              uses: actions/setup-python@3105fb18c05ddd93efea5f9e0bef7a03a6e9e7df
+              with:
+                  python-version: "3.9.x"
 
-      - name: Install Poetry
-        uses: Gr1N/setup-poetry@462ac83c852d49e282a1233c4c24c5411696e7c7
+            - name: Get virtual env path
+              id: venv-path
+              run: |
+                  poetry env use python
+                  echo "::set-output name=dir::$(poetry env info -p)"
+              working-directory: ${{ env.SRC_DIR }}
 
-      - name: Get virtual env path
-        id: venv-path
-        run: |
-            poetry env use python
-            echo "::set-output name=dir::$(poetry env info -p)"
-        working-directory: ${{ env.SRC_DIR }}
+            - name: Setup Dependency Caching
+              uses: actions/cache@c64c572235d810460d0d6876e9c705ad5002b353
+              id: cache-venv
+              with:
+                  path: ${{ steps.venv-path.outputs.dir }}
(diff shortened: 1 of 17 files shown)


src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
```diff
@@ -0,0 +1,79 @@
...
+    Parameters
+    ----------
+    canvas: Canvas
+        Canvas object.
+    start_date: str
+        Retrieve Courses starting on or after this date.
+    end_date:
+        Retrieve Courses ending on or before this date.
+    sync_db: sqlalchemy.engine.base.Engine
+        Sync database connection.
```
stephenfuqua, line 44: Parameters block does not align with actual function parameters. I assume you'll review all of the document blocks before the final code review; therefore I won't comment on any others.

src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
```diff
@@ -0,0 +1,79 @@
...
+from edfi_canvas_extractor.mapping import (
+    sections as sectionsMap,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+def extract_courses(
+    gql: Extract,
```
stephenfuqua, line 29: `Extract` is one of those strange English words that is both a verb and a noun. It took me a little while to understand what I was seeing, because initially I assumed that `Extract` is the _result_, that is, the X in "I extracted X". As a noun this didn't make sense to me. Once I thought of it as a verb, it made more sense! Suggestion: rename `Extract` to `GraphQLExtractor` or something like that. "Extractor" makes more clear that this is the module that "extracts from GraphQL".

src/canvas-extractor/edfi_canvas_extractor/extract_graphql.py
```diff
@@ -0,0 +1,103 @@
...
+
+    succeeded = _get_courses(gql, sync_db)
+
+    if not succeeded:
+        _break_execution("Courses")
+
+    succeeded = _get_sections(arguments, gql, sync_db)
+
+    if not succeeded:
+        _break_execution("Sections")
```
stephenfuqua, line 65: This function is now doing three things: getting accounts, getting courses, and getting sections. The name implies that it is only taking the first of these actions.

src/canvas-extractor/edfi_canvas_extractor/graphql/canvas.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __getattribute__(self, name):
+        return super(CanvasObject, self).__getattribute__(name)
+
+    def __init__(self, requester, attributes):
+        """
+        :param requester: The requester to pass HTTP requests through.
+        :type requester: :class:`canvasapi.requester.Requester`
+        :param attributes: The JSON object to build this object with.
+        :type attributes: dict
```
stephenfuqua, line 26: Please use [NumPy-style docstrings](https://numpydoc.readthedocs.io/en/v1.4.0/format.html).

src/canvas-extractor/edfi_canvas_extractor/graphql/canvas.py
```diff
@@ -0,0 +1,76 @@
...
+    def __repr__(self):  # pragma: no cover
+        classname = self.__class__.__name__
+        attrs = ", ".join(
+            [
+                "{}={}".format(attr, val)
+                for attr, val in self.__dict__.items()
+                if attr != "attributes"
+            ]
+        )  # noqa
+        return "{}({})".format(classname, attrs)
```
stephenfuqua, line 40: As a mild style preference, we've been using string interpolation almost everywhere. There was some situation with exceptions where we chose not to use it, though I can't remember the detail of why now.
```suggestion
        return f"{classname}({attrs})"
```

stephenfuqua (commented):
Thank you for sharing this draft with me. Seems that you're making good progress on the solution now. I realize that it is not complete. I did not read thoroughly or comment on everything, since you'll probably go through the code again on your own before submitting the final review.

src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
sgonzalezgap: Got the same issue naming variables, etc. even on spanish so I am ok with that name of class, Working on ...

src/canvas-extractor/edfi_canvas_extractor/graphql/extractor.py
```diff
@@ -0,0 +1,181 @@
...
+            query string for GraphQL
+
+        Returns
+        -------
+        Dict JSON Object
+        """
+
+        load_dotenv()
+        CANVAS_URL = os.getenv('CANVAS_BASE_URL')
+        CANVAS_TOKEN = os.getenv('CANVAS_ACCESS_TOKEN')
```
stephenfuqua, line 56: These values are parsed from the command line at runtime and should be passed as arguments, rather than as environment variables. See the `MainArguments` object passed to the `run()` function.

src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
```diff
@@ -26,20 +26,16 @@
...
     sync_db: sqlalchemy.engine.base.Engine,
 ) -> Tuple[List[Course], DataFrame]:
     """
-    Gets all Canvas courses for the given date range, in the Ed-Fi UDM format.
+    Gets all Canvas courses in the Ed-Fi UDM format.
 
     Parameters
     ----------
     canvas: Canvas
         Canvas object.
```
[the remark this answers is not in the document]
sgonzalezgap: Not quite sure about this one, could you please clarify? 
Its a substitution in the docstring?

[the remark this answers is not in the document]
stephenfuqua: `canvas` was the old argument, you replaced it with the `gql` argument, but forgot to update the docstring to have the new parameter information.

stephenfuqua:
Something is going wrong with the CLA process, perhaps it is because the PR is in draft? I can see that Sergio has signed the agreement;
image

stephenfuqua:
@salfredogonzalez please be sure to address the [failing tests](https://github.com/Ed-Fi-Alliance-OSS/LMS-Toolkit/actions/runs/3055729673/jobs/4929114578)

.github/workflows/test-canvas.yml
```diff
@@ -5,61 +5,61 @@
...
-        with:
-          # This one needs to be hard-coded to avoid 3.9.5, due to a bug
-          # in one of the dependencies
-          python-version: '3.9.4'
+            - name: Install Python 3.9
+              uses: actions/setup-python@3105fb18c05ddd93efea5f9e0bef7a03a6e9e7df
+              with:
+                  # This one needs to be hard-coded to avoid 3.9.5, due to a bug
+                  # in one of the dependencies
+                  python-version: "3.9.4"
```
stephenfuqua, line 37: If you move "Install Python" after "Install Poetry", then you can setup Poetry caching on the "Install Python" step and then remove the "get virtual env path" and "setup dependency caching" steps.

.github/workflows/test-sql-adapter.yml
```diff
@@ -5,59 +5,62 @@
```
stephenfuqua, line 5: Can you help me understand why this file changed? Perhaps it should be reverted. I know we'll need to update the Poetry install in the future, but perhaps that should be in a separate ticket.

I have now created a new ticket for updating _all_ of the Actions.

src/canvas-extractor/pyproject.toml
```diff
@@ -1,39 +1,40 @@
...
-twine = "^3.3.0"
-Faker = "^6.5.0"
-pyfakefs = "^4.5.6"
-
-[build-system]
-requires = ["poetry>=0.12"]
-build-backend = "poetry.masonry.api"
+[tool.poetry]
+name = "edfi-canvas-extractor"
+version = "1.2.1"
```
stephenfuqua, line 3:
```suggestion
version = "2.0.0"
```

PR #273, opened 2022-09-26
[LMS-470] Rewrite Canvas Extractor to use GraphQL for Assignments

sgonzalezgap:
Adding the functionality to extract assignments with GraphQL


src/canvas-extractor/edfi_canvas_extractor/extract_graphql.py
```diff
@@ -126,6 +137,26 @@ def _write_sections_associations(arguments: MainArguments) -> None:
...
+    sync_db: sqlalchemy.engine.base.Engine,
+) -> bool:
+    logger.info("Extracting Assignments from Canvas")
+    (sections, _, all_section_ids) = results_store["sections"]
+    sections_df = to_df(sections)
+    (assignments, udm_assignments_df) = extract_assignments(
+        sections_df, gql, sync_db
+    )
+    logger.info("Writing LMS UDM Assignments to CSV files")
+    write_assignments(
```
saa14, line 152: SUGGESTION: This can be separated out into it's own function `_write_assignments` with the `@catch_exceptions` decorator like we have for other similar operations in the file.

src/canvas-extractor/tests/graphql/assignments/test_extractor_assignments.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_assignments(gql_no_run, api):
+    """
+    Get from the sample data
+    obtain the assignments info
+    Check and check the return type
+    """
+    gql_no_run.extract(api)
+    assignments = gql_no_run.assignments
```
saa14, line 18: Here and in similar places in other tests for `assignments, courses, enrollments, sections and students` , we directly use `gql_no_run.assignments`, shouldn't we use the `get_assignments` functions in the  `GraphQLExtractor` class.

src/canvas-extractor/tests/graphql/extractor/test_extractor_datefilter.py
```diff
@@ -0,0 +1,40 @@
...
+            continue
+
+        start_term = courses["term"]["startAt"]
+        end_term = courses["term"]["endAt"]
+
+        assert start_term is None
+        assert end_term is None
+
+
+@pytest.mark.xfail
```
saa14, line 28: COULD: Instead of marking this test (and the next one) as `expected to fail` , we could raise a `ValueError` exception in `validate_date()` and assert that error in these tests.

src/canvas-extractor/tests/graphql/extractor/test_extractor_datefilter.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.mark.parametrize("date", ["", None, False])
+def test_validate_date_value(date):
+    assert date is True
```
saa14, line 31: SHOULD: This test does not seem to assert anything useful. Did you mean to use `validate_date(date)` here?

sgonzalezgap: I probably need to remove this one, at the moment helps me to validate the type that I was receiving from the API.

saa14:
@sgonzalezgap Thanks for making the changes. As discussed we will be adding additional tests (against the GraphQL Canvas API) in a [future ticket](https://tracker.ed-fi.org/browse/LMS-488).

PR #276, opened 2022-09-28
[LMS-469] Extract enrollments and students from GraphQL

sgonzalezgap:


src/canvas-extractor/edfi_canvas_extractor/graphql/students.py
```diff
@@ -0,0 +1,77 @@
...
+        a list of Canvas Users objects
+    sync_db: sqlalchemy.engine.base.Engine
+        an Engine instance for creating database connections
+
+    Returns
+    -------
+    DataFrame
+        a Students API DataFrame with the current and previously fetched data
+    """
+    students_df: DataFrame = _sync_without_cleanup(to_df(students), sync_db)
```
stephenfuqua, line 46: Comparing the Student list created by the old extractor and your update, there are four extra students now. Apparently we had a bug before that we didn't know about! And now the bug is fixed 😄 .

src/canvas-extractor/edfi_canvas_extractor/graphql/extractor.py
```diff
@@ -45,8 +73,8 @@ def get_from_canvas(self, query: str):
         Dict JSON Object
         """
 
-        GRAPHQL_URL = f"{self.base_url}/api/graphql"
-        GRAPHQL_AUTH = {'Authorization': f'Bearer {self.access_token}'}
+        GRAPHQL_URL = f"{self.url}/api/graphql"
+        GRAPHQL_AUTH = {'Authorization': f'Bearer {self.token}'}
 
         try:
             fetch = requests.post(
```
stephenfuqua, line 80: I just ran the GraphQL query in Postman, with page size of only 20. I got a gateway 504 error. Then tried again a few seconds later. It completed, but it took 48.67 seconds to run. 

We might need to add a parameter to control the timeout setting on the request. This should probably be a separate ticket though.

src/canvas-extractor/edfi_canvas_extractor/graphql/extractor.py
```diff
@@ -105,23 +135,81 @@ def extract(self, body) -> None:
...
+                        "sis_user_id": users["sisId"],
+                        "created_at": users["createdAt"],
+                        "name": users["name"],
+                        "email": users["email"],
+                        "login_id": users["loginId"],
+                    })
+
+                    self.enrollments.append({
+                        "id": enrollment["_id"],
+                        "user_id": users["name"],
```
stephenfuqua, line 153: Small bug here. Should be:
```suggestion
                        "user_id": users["_id"],
```

src/canvas-extractor/edfi_canvas_extractor/extract_graphql.py
```diff
@@ -75,10 +109,13 @@ def run(arguments: MainArguments) -> None:
 
     for account in accounts:
         _id = getattr(account, "id")
```
stephenfuqua, line 111: Let's add a logging statement to help us know which accounts are available:

```suggestion
        _id = getattr(account, "id")
        logger.info(f"Getting data for account ID {_id}")
```

src/canvas-extractor/edfi_canvas_extractor/extract_graphql.py
```diff
@@ -75,10 +109,13 @@ def run(arguments: MainArguments) -> None:
...
-        gql.set_account(_id)
-        gql.set_dates(arguments.start_date, arguments.end_date)
+        gql = GraphQLExtractor(
+            arguments.base_url,
+            arguments.access_token,
+            _id,
+            arguments.start_date,
+            arguments.end_date
+            )
         gql.run()
```
stephenfuqua, line 119: We are not getting any files for account_id = 2. There is a scope problem: note that `gql` is instantiated on line 112, inside the `for` loop. However, we then use `gql` _outside_ the for loop, starting at line 124. Thus we end up only getting the data from the last account in the list from `get_accounts`.

stephenfuqua (changes requested):

src/canvas-extractor/edfi_canvas_extractor/graphql/extractor.py
stephenfuqua: https://stackoverflow.com/questions/38965086/why-is-default-timeout-for-python-requests-get-different-for-different-machine

I have seen a couple of other articles mentioning this too: there is no default timeout setting. So this will just run until there is completion or a gateway timeout. That's fine with me. Therefore no new ticket is needed at this time, and we will not add a timeout setting to the `request.post` command.

src/canvas-extractor/edfi_canvas_extractor/extract_graphql.py
sgonzalezgap: applied changes

saa14:
Overall, the changes look great, and the `GraphQL` results appear to have the same section-associations as seen in `REST API` results. A few queries/suggestions:
- QUERY: I noticed we use the `canvasapi` package for `CanvasObject` class in the `api` module. However, we created a new similar class in  `graphql\canvas_object.py` and then used it in `graphql\canvas_helper.py`. I think the one coming from the package is not much different from what we have in the `graphql` code. Am I missing something?
- QUERY: Are we planning on adding unit tests for GraphQL like we have for the rest api code in a separate ticket? Or should we include it (or modify existing tests) as we go along these tickets?
- SUGGESTION: More descriptive commit messages and smaller multiple commits with a limited scope would be helpful for code reviews/future work as the Tools team is still a bit new to the codebase.

src/canvas-extractor/edfi_canvas_extractor/graphql/extractor.py
sgonzalezgap: Changes applied

src/canvas-extractor/edfi_canvas_extractor/graphql/extractor.py
```diff
@@ -4,33 +4,53 @@
...
-        self.data = None
+        Parameters
+        -----------
+        url: str
+            Base URL to connect to Canvas
+        token: str
+            Secret to get access to Canvas
+        """
         self.courses = list()
+        self.data = False
```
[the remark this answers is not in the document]
stephenfuqua: I like this suggestion.

stephenfuqua:
> * QUERY: I noticed we use the `canvasapi` package for `CanvasObject` class in the `api` module. However, we created a new similar class in  `graphql\canvas_object.py` and then used it in `graphql\canvas_helper.py`. I think the one coming from the package is not much different from what we have in the `graphql` code. Am I missing something?

I think there is a transition in process away from the `canvasapi` package to our own code. I agree that we don't need to use `canvasapi` just for plain objects, if doing so makes it more difficult to map the GraphQL results into objects that we can use.

> * QUERY: Are we planning on adding unit tests for GraphQL like we have for the rest api code in a separate ticket? Or should we include it (or modify existing tests) as we go along these tickets?

I should have noticed that with the first ticket :-). For LMS-469, let's try to move forward without additional tests. But for the next stories, @sgonzalezgap please start adding unit tests.

> * SUGGESTION: More descriptive commit messages and smaller multiple commits with a limited scope would be helpful for code reviews/future work as the Tools team is still a bit new to the codebase.

Yes, I agree please 😁 .

saa14 (approved):

PR #277, opened 2022-10-07
[LMS-471] Extract assignment submissions from Canvas

sgonzalezgap:


src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
```diff
@@ -9,23 +9,22 @@
 from pandas import DataFrame
 from typing import Dict, List, Tuple
 
-from canvasapi.course import Course
-from canvasapi.enrollment import Enrollment
-from canvasapi.section import Section
-from canvasapi.user import User
+from canvasapi.submission import Submission
```
saa14, line 12: QUESTION: Is there a reason we are still using `canvasapi` for `Submission` instead of `Dict[str, str]`?

src/canvas-extractor/edfi_canvas_extractor/extract_graphql.py
```diff
@@ -165,6 +167,22 @@ def _write_assignments(
...
 
+@catch_exceptions
+def _get_submissions(
+    arguments: MainArguments,
+    gql: GraphQLExtractor,
+    sync_db: sqlalchemy.engine.base.Engine,
+) -> None:
+    logger.info("Extracting Submissions from Canvas")
+    (sections, _, _) = results_store["sections"]
+    logger.info("Writing LMS UDM AssignmentSubmissions to CSV files")
```
saa14, line 178: SHOULD: This can be separated out into it's own function `_write_submissions` with the `@catch_exceptions` decorator like we have for other similar operations in the file.

src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
sgonzalezgap: Remove all the canvasapi dependecies

src/canvas-extractor/edfi_canvas_extractor/extract_graphql.py
sgonzalezgap: There is a subject here, when its on separate functions stops the mapping. It could be the way that its passing the params to the write function in mapping. Meanwhile I leave it working.

src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
```diff
@@ -193,3 +190,96 @@ def extract_assignments(
...
+    -------
+    Dict[Tuple[str, str], DataFrame]
+        A dict with (section_id, assignment_id) as key and udm_submissions
+        as value.
+    """
+    export: Dict[Tuple[str, str], DataFrame] = {}
+    submissions: List[Dict[str, str]] = gql.get_submissions()
+    if len(list(submissions)) < 1:
+        logger.info(
+            "Skipping submissions for section id %s - No data returned by API",
```
saa14, line 218: SHOULD: We have to either provide the section id here or reformat this message in this log message to not use `%s`.

src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
```diff
@@ -193,3 +190,96 @@ def extract_assignments(
...
+        for assignment_id, submissions_df in submissions_dfs_by_assignment_id.items():
+            section_id = str(submissions_df.iloc[0]['section_id'])
+            assignment_source_system_identifier = f"{section_id}-{str(assignment_id)}"
+            submissions_df["assignment_id"] = assignment_source_system_identifier
+            submissions_df = submissionsMap.map_to_udm_submissions(submissions_df, section_id)
+            export[(section_id, f"{assignment_source_system_identifier}")] = submissions_df
+        return export  # type: ignore
+
+
+# def extract_submissions(
```
saa14, line 240: COULD: If this version of `extract_submissions` is not intended to be used or expected to be re-introduced in a near future PR, I would prefer we remove it and rely on the repo history to get it back if needed.

sgonzalezgap: Miss that one, removing ...

src/canvas-extractor/edfi_canvas_extractor/client_graphql.py
sgonzalezgap: SUGGESTION:  "Skipping submissions for section - No data returned by API" ?

saa14 (approved):

PR #279, opened 2022-10-20
[LMS-488] Integration test

sgonzalezgap:
Adding unit testing and integration testing


src/canvas-extractor/tests/graphql/extractor/test_extractor_integration.py
```diff
@@ -0,0 +1,179 @@
...
+        "1",
+        START_DATE,
+        END_DATE,
+        )
+    yield gql
+
+
+@pytest.fixture(scope="class")
+def graphql_class(gql_class):
+    query = query_builder(account_id=1)
```
saa14, line 42: SHOULD: Instead of building the query here and then calling `get_from_canvas`, we should use use `gql_class.run()` which performs the same operations and also sets `has_data` flag which can then be asserted in the test.

src/canvas-extractor/tests/graphql/extractor/test_extractor_integration.py
```diff
@@ -0,0 +1,179 @@
...
+    DB_FILE = "tests/graphql/test.db"
+    Path(DB_FILE).unlink(missing_ok=True)
+    yield create_engine(f"sqlite:///{DB_FILE}", echo=True)
+    # Path(DB_FILE).unlink(missing_ok=True)
+
+
+@pytest.mark.skip
+class TestExtractorIntegration:
+
+    def test_extractor_request(self, gql_class):
```
saa14, line 60: SHOULD: We can refactor this to just assert the `has_data` flag instead of performing the query building and fetching from canvas operations again.

src/canvas-extractor/tests/graphql/extractor/test_extractor_integration.py
```diff
@@ -0,0 +1,179 @@
...
+        CANVAS_ACCESS_TOKEN,
+        "1",
+        START_DATE,
+        END_DATE,
+        )
+    yield gql
+
+
+@pytest.fixture(scope="class")
+def graphql_class(gql_class):
```
saa14, line 41: SHOULD: We should move towards having only a single fetch request for the entire test class instead of passing in this fixture to each test function in the class. We can make use of the concept of [autouse fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) and [request fixture](https://docs.pytest.org/en/stable/reference/reference.html#request) to achieve this. Something like this:
```
@pytest.fixture(autouse=True, scope="class")
def fetch_graphql_data(request, gql_run):
    request.cls._graphql_extraction = gql_run
```
The tests can then use assertions as follows:
```
...
def test_gql_data_fetched(self):
        assert self._graphql_extraction.has_data is True

def test_gql_courses_not_empty(self):
        courses = self._graphql_extraction.get_courses()
        assert courses is not None
        assert isinstance(courses, list)
...
```

saa14:
SHOULD: We should add a way to locally run the `test_extractor_integration` tests. We can maybe add `load_dotenv` in `conftest.py` and provide guidance to either add a local `.env` file for tests in the same directory or provide the path to a local `.env` file.

src/canvas-extractor/tests/graphql/extractor/test_extractor_integration.py
sgonzalezgap: The first test it's checking that is getting information before the run method, just to be clear that will remove the graphql_class and leave gql_class to add the run method and change test_extractor_request to assert if has data, correct?

saa14: Yes. The idea is to only run the fetch request once and perform the assertions on the already fetched data. The below [comment](https://github.com/Ed-Fi-Alliance-OSS/LMS-Toolkit/pull/279#discussion_r1022900842) makes it clearer. In this case, I think the `has_data` flag is a good indicator if we actually get any data or not.

src/canvas-extractor/tests/graphql/conftest.py
```diff
@@ -34,19 +47,33 @@ def gql(api):
...
+        END_DATE,
+        )
+    gql.run()
+
+    yield gql
+
+
+@pytest.fixture(autouse=True, scope="class")
+def load_env():
+    load_dotenv()
```
saa14, line 76: SHOULD: `load_dotenv()` can be directly called without wrapping it as a fixture.

saa14 (approved):
Thanks for addressing the feedback. The changes look great.

PR #283, opened 2022-11-10
[LMS-485] Update all Python Setups in GitHub Actions

sgonzalezgap:


.github/workflows/publish-extractor-lib.yml
```diff
@@ -19,14 +19,14 @@ jobs:
...
         uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
 
+      - name: Install Poetry
+        run: pipx install poetry
+
       - name: Install Python 3.9
-        uses: actions/setup-python@3105fb18c05ddd93efea5f9e0bef7a03a6e9e7df
+        uses: actions/setup-python@98f2ad02fd48d057ee3b4d4f66525b231c3e52b6 # v3.1.2
         with:
           python-version: '3.9.x'
```
saa14, line 28: SHOULD: The `cache` input is missing here.

saa14:
- The workflows seem to still have the following steps:
```
      - name: Get virtual env path
        id: venv-path
        run: |
            poetry env use python
            echo "::set-output name=dir::$(poetry env info -p)"
        working-directory: ${{ env.SRC_DIR }}
      - name: Setup Dependency Caching
        uses: actions/cache@c64c572235d810460d0d6876e9c705ad5002b353
        id: cache-venv
        with:
          path: ${{ steps.venv-path.outputs.dir }}
          key: ${{ runner.os }}-poetry-${{ hashFiles(env.LOCK_FILE) }}
          restore-keys: |
            ${{ runner.os }}-poetry-
```
These need to removed in favor of dependency caching within the `Install Python` task using the `cache` input with `poetry`. 

- Some workflows also seem to have duplicate `Install Poetry` tasks which can be removed in favor of the new ones.
```
      - name: Install Poetry
        uses: Gr1N/setup-poetry@462ac83c852d49e282a1233c4c24c5411696e7c7
```

.github/workflows/publish-extractor-lib.yml
sgonzalezgap: Do you mean I should add 

`- name: Setup Dependency Caching
        uses: actions/cache@c64c572235d810460d0d6876e9c705ad5002b353
        id: cache-venv
        with:
          path: ${{ steps.venv-path.outputs.dir }}
          key: ${{ runner.os }}-poetry-${{ hashFiles(env.LOCK_FILE) }}
          restore-keys: |
            ${{ runner.os }}-poetry-`

Below this line in the code?

stephenfuqua:
@sgonzalezgap my apologies, I ended up doing this in my PR as well, after forgetting that there was a separate ticket for just updating the action versions.

saa14:
The changes intended in this PR have been covered in the [broader workflow revamp PR](https://github.com/Ed-Fi-Alliance-OSS/LMS-Toolkit/pull/286). So, we are closing this one.

PR #284, opened 2022-11-11
[LMS-480] Notebook instructions point to wrong README file

sgonzalezgap:


saa14:
> This PR should be merging into `main` instead of `LMS-308-canvas-graphql` as it is not part of the canvas rewrite epic.

Also, this branch should be based on top of `main`. Currently, it seems to be based on `LMS-308-canvas-graphql`

docs/demonstration/google-e2e.ipynb
```diff
@@ -261,7 +261,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3.9.4 64-bit ('edfi')",
```
saa14, line 264: QUERY: Are the updates to `display_name` and `vscode` intended?

docs/demonstration/schoology-e2e.ipynb
```diff
@@ -1613,7 +1613,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3.9.4 64-bit ('edfi')",
```
saa14, line 1616: QUERY: Are the updates to `display_name` and `vscode` intended?

stephenfuqua: These updates were made by the tool, and I don't see any harm in them. You can double-check that it works fine still from `poetry run jupyter notebook` - I'm quite confident that it won't be a problem.

docs/demonstration/schoology-e2e.ipynb
```diff
@@ -18,7 +18,7 @@
     "## Getting Started\n",
     "\n",
     "1. Confirm you have [poetry](https://python-poetry.org) installed (`poetry --version`).\n",
-    "1. Follow the [notebook instructions](README.md) to install dependencies used by this notebook.\n",
+    "1. Follow the [notebook instructions](../../src/schoology-extractor/README.md) to install dependencies used by this notebook.\n",
```
stephenfuqua, line 21: The original instructions are for installing the notebook's dependencies, not Schoology Extractor's dependencies. So I don't think this change is correct.

docs/demonstration/canvas-e2e.ipynb
```diff
@@ -16,7 +16,7 @@
     "## Getting Started\n",
     "\n",
     "1. Confirm you have [poetry](https://python-poetry.org) installed (`poetry --version`).\n",
-    "1. Follow the [notebook instructions](README.md) to install dependencies used by this notebook.\n",
+    "1. Follow the [notebook instructions](../../src/canvas-extractor/README.md) to install dependencies used by this notebook.\n",
```
stephenfuqua, line 19: I don't think this change is correct.

stephenfuqua: For LMS-480: there are [dependencies](https://github.com/Ed-Fi-Alliance-OSS/LMS-Toolkit/blob/main/src/notebooks/pyproject.toml) for the notebook folder itself, so the original line was correct - you need to install the dependencies for the notebook folder. Think about reading this file in GitHub or in VS Code - you can be reading it without having launched Jupyter Notebook.

But I think you're right that another line is needed: a reminder to install the extractor's dependencies.

docs/demonstration/google-e2e.ipynb
saa14: Agreed with the reasoning [here](https://github.com/Ed-Fi-Alliance-OSS/LMS-Toolkit/pull/284#discussion_r1026567842)

saa14 (approved):

PR #288, opened 2022-11-23
[LMS-491] Add pytest markers for unit and integration tests

sgonzalezgap:


saa14:
SHOULD: The markers are being registered twice with `pytest.ini` and the changes in `conftest.py`. We should only keep one of them. I think `pytest.ini` can be brought out to the `src/canvas_extractor` folder to achieve that and remove `conftest.py` changes then. Or we can keep the `conftest.py` changes and remove `pytest.ini`.

saa14 (approved):
The changes look great and allows us to run `integration` and `unit` tests separately based on the context. Future tickets would focus on:
- Updating other tests in the project with the markers.
- Updating to run the tests using markers in `github-action` workflows

PR #289, opened 2022-11-29
Revert "[LMS-480] Notebook instructions point to wrong README file"

saa14:
Reverts Ed-Fi-Alliance-OSS/LMS-Toolkit#284
**NOTE**
The demonstration notebooks pull the extractors from [pypi.org](http://pypi.org/), not from the local repository. In other words, they were installed using `poetry add`. Thus poetry handles the dependency installation, and there is no need to install the dependencies from the extractor source code locations.


sgonzalezgap (approved):
Reverting on Stephen request

PR #294, opened 2022-12-14
GraphQL and other updates

stephenfuqua:


src/canvas-extractor/tests/graphql/integration/test_extractor_integration.py
```diff
@@ -32,9 +32,6 @@ def graphql(request, gql):
 @pytest.mark.usefixtures("graphql")
 class TestExtractorIntegration:
 
-    def test_gql_data_fetched(self):
-        assert self.graphql_extractor.has_data is True  # type: ignore
-
```
stephenfuqua, line 37: This is an invalid test: the `gql` fixture runs `gql.extract(api)`, which never sets `has_data = True`. The test is only valid if the fixture is changed to call `gql.run()`. All of the other tests are passing in the current approach, so I just removed this test above

PR #395, opened 2025-03-07
[DOC-197] Dependency and Link Updates

stephenfuqua:
This was harder than expected - so much tech debt after about 2 years!

* Updated GitHub Actions
* Worked through all dependency vulnerabilities
* All tests are passing
* Had to fix some Python type problems detected by updated linting tools


jasonh-edfi (approved):
Reviewed and approved for version updates.

